PR #163, opened 2023-03-15
Open sourcing

patch0:
Adds:
* Licence and copyright
* Changelog
* Contributing guidance
* Docs in the readme


sra405:
Didn't realise this was an open PR! Thankfully mostly the same as what we ended up uses except the license 😄

PR #225, opened 2023-11-14
Tweaks to allow using another database in local development

ce-rpf:
Three changes here:

* import `POSTGRES_HOST` from the environment (`.env` file) rather than presetting it in docker-compose; this makes it easier to use another database without editing version-controlled files). This also brings editor-api into consistency with how projects-admin does it
* move postgres database exposed port to 5434, to avoid a clash with projects-admin
* add the shared development network so editor-api can see postgres/nginx in core-dev-services for those who choose to use it

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       - POSTGRES_PASSWORD
       - POSTGRES_USER
     ports:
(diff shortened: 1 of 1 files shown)


PetarSimonovic (approved):
lgtm 👍

PR #234, opened 2024-02-08
Add a User model to support the Learning Management MVP

tuzz:
Previously, the editor-api set the current_user to the ID of a user that was returned from the HydraAdminAPI. To support requirements relating to the Learning Management MVP, we need more information about users such as their roles (school-owner, school-teacher, school-student) and their names so that we can present pages that contain lists of students within a class.

This data is not returned from the HydraAdminAPI so we need to switch to using the HydraPublicAPI that does provide this information for the currently logged in user. To retrieve the information of other users, we can integrate with the UserInfoAPI which allows you to pass a list of user UUIDs and retrieve their profile information.

This pull request introduces a User model that can be initialized with `User.from_omniauth(token: "<access token>")` or `User.where(id: [<list of user UUIDs>])`. Instances can then be used in a Rails-like way, i.e. `user.name`, `user.email`, etc.

This is a large refactor since it touches all parts of the code that depend on the current_user, such as ApiController, GraphqlController, CanCanCan's Ability model and a large number of tests. I have replaced the HydraAdminApiMock with HydraPublicApiMock which provides some helpful utility methods to be used in tests.

This pull request depends on the following new environment variables:
- USERINFO_API_URL
- USERINFO_API_KEY
- HYDRA_PUBLIC_URL

And it no longer depends on:
- HYDRA_ADMIN_URL
- HYDRA_ADMIN_API_KEY

We will need to ensure these environment variables have been updated in staging/production before this work can be deployed.

Many of the files in this pull request were copied from [clubs-api](https://github.com/RaspberryPiFoundation/clubs-api) and some were adapted slightly. For example, the `User` model and its associated spec and the two API clients.


app/jobs/upload_job.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'open-uri'
-require 'project_importer'
```
sra405, line 4: ❓ how come this has been removed?

sra405: aha! Nice!

sra405 (approved):
👍 nice work, looks and works ace locally

PR #261, opened 2024-04-17
Do not load unused Rails frameworks

floehopper:
Since we're not using ActionCable, ActionText, ActionMailer or ActionMailbox, we can simplify the app by not loading them and removing related configuration and files.


chrislo (approved):
Looks good to me. I think there's some upcoming requirements to send emails as part of the lerning management work, but I don't think that should stop us removing ActionMailer for now, we can always add it back if/when we need it.

PR #262, opened 2024-04-17
Upgrade Ruby from v3.2.1 -> v3.2.3

floehopper:
This is the latest patch version of Ruby v3.2.

We've been seeing the following warning in our Heroku deployments, e.g. in [this staging deployment][1]:

    There is a more recent Ruby version available for you to use:

    3.2.3

    The latest version will include security and bug fixes. We always recommend
    running the latest version of your minor release.

    Please upgrade your Ruby version.

    For all available Ruby versions see:
     https://devcenter.heroku.com/articles/ruby-support#supported-runtimes

The Ruby version in `Gemfile` and the base image in `Dockerfile` are only specified in terms of the minor version (i.e. v3.2) rather than the patch version, so there's no need to change them. Furthermore the `ruby:3.2-slim-bullseye` base image has already been [updated to include v3.2.3][2].

[1]: https://dashboard.heroku.com/apps/editor-api-staging/activity/builds/f9360560-9d26-487b-8258-cf84906055bf
[2]: https://github.com/docker-library/ruby/commit/c473741514de6aa1e3ccdf3a9c6df0aa71348ce3


floehopper:
I've checked the output for the Heroku build for this branch and I can see that the "There is a more recent Ruby version available for you to use" warning no longer appears and it is using Ruby v3.2.3 OK.

chrislo (approved):

PR #266, opened 2024-04-18
Fix spec for projects:create_starter Rake task

floehopper:
The `spec/tasks/create_starter_spec.rb` was triggering a `NameError` exception when run on its own (using `bundle exec rspec spec/tasks/create_starter_spec.rb`) and when the whole test suite was run in a certain order (e.g. `bundle exec rspec --seed 54367`). The main goal of this PR is to fix that problem.

I've achieved that by extracting all the code out of the Rake task and into a new `FilesystemProject.import_all!` method and testing that directly rather than testing the Rake task which tends to cause class loading problems.

As part of doing this, I've also removed the stubbing from the spec, because I found it confusing and it turns out it doesn't take too long to run import all the projects for real in the spec.

See individual commit notes for more details.


danhalson (approved):
Looks good, nice refactor of the phrase seed too

PR #267, opened 2024-04-19
Update school fields

danhalson:
Ready for review

Adds missing fields and tweaks model validations

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,15 +16,15 @@ USERINFO_API_KEY=1234
 
 HYDRA_PUBLIC_URL=http://localhost:9001
 HYDRA_PUBLIC_TOKEN_URL=http://host.docker.internal:9001
-HYDRA_CLIENT_ID=editor-api-dev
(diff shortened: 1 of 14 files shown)


.env.example
```diff
@@ -16,15 +16,15 @@ USERINFO_API_KEY=1234
 
 HYDRA_PUBLIC_URL=http://localhost:9001
 HYDRA_PUBLIC_TOKEN_URL=http://host.docker.internal:9001
-HYDRA_CLIENT_ID=editor-api-dev
+HYDRA_CLIENT_ID=editor-dashboard-dev
```
loiswells97, line 19: More a question for my understanding than anything else, but why does this need to change?

danhalson: Scott added a new role on profile to control access to the editor dashboard, without using this config you won't be able to login locally: https://github.com/RaspberryPiFoundation/profile/pull/1563

loiswells97 (approved):
Looks good to me 👍

Issue #244, opened 2024-02-23
Determine if we handle app errors correctly outside local dev


sra405:
Currently there's a block that appears to only consider local requests before categorising the requests correctly. I believe this results in 500s being returned by the app in staging/prod. Needs investigating.

♻️ I think you're right about this block, I'm pretty sure we only get 500s from the app in production if these cases occur. I'll backlog it 👍

_Originally posted by @sra405 in https://github.com/RaspberryPiFoundation/editor-api/pull/233#discussion_r1500716950_

chrisroos:
> Currently there's a block that appears to only consider local requests before categorising the requests correctly. I believe this results in 500s being returned by the app in staging/prod. Needs investigating.

It's actually the other way around. [The block](https://github.com/RaspberryPiFoundation/editor-api/blob/fce9e4e33d3ca697501d2e75f3cc8d627c7db4bf/app/controllers/api_controller.rb#L8-L13) only categorises the errors correctly when `consider_all_requests_local` is false (i.e. test and production envs). I've opened a PR to remove the condition in https://github.com/RaspberryPiFoundation/editor-api/pull/275 so that we see consistent behaviour in all environments.

PR #270, opened 2024-04-22
update school validations to return list of errors for frontend use

conorriches:
### Issue
Related to https://github.com/RaspberryPiFoundation/editor-standalone/issues/28

### What changed
When a school is trying to be created, but validations fail, we now get this response:
```
{
  "errors": {
    "website":["can't be blank"],
    "address_line_1":["can't be blank"],
    "municipality":["can't be blank"],
    "country_code":["can't be blank","is not included in the list"]
  }
}
```

This is useful, as the FE application can now iterate on the keys, and selectively update each input field to show the error message.

Before this change we got:
```
{"error":"Error creating school: Website can't be blank,Address line 1 can't be blank,Municipality can't be blank,Country code can't be blank,Country code is not included in the list"}
```

which isn't useful, as we can't associate errors to fields

### DIFF
--- a/app/controllers/api/schools_controller.rb
+++ b/app/controllers/api/schools_controller.rb
@@ -15,13 +15,14 @@ module Api
     end
 
     def create
-      result = School::Create.call(school_params:)
+      @school = School.new(school_params)
 
-      if result.success?
-        @school = result[:school]
+      if @school.save
         render :show, formats: [:json], status: :created
       else
-        render json: { error: result[:error] }, status: :unprocessable_entity
+        render json: {
+          errors: @school.errors
+        }, status: :unprocessable_entity
       end
     end
 
--- a/app/controllers/api/schools_controller.rb
+++ b/app/controllers/api/schools_controller.rb
@@ -15,14 +15,13 @@ module Api
     end
 
     def create
-      @school = School.new(school_params)
+      result = School::Create.call(school_params:)
 
-      if @school.save
+      if result.success?
+        @school = result[:school]
         render :show, formats: [:json], status: :created
       else
-        render json: {
-          errors: @school.errors
-        }, status: :unprocessable_entity
+        render json: { error: result[:error] }, status: :unprocessable_entity
       end
     end
 
--- a/spec/concepts/school/create_spec.rb
+++ b/spec/concepts/school/create_spec.rb
@@ -55,9 +55,14 @@ RSpec.describe School::Create, type: :unit do
       expect(response.failure?).to be(true)
     end
 
-    it 'returns the error message in the operation response' do
+    it 'returns the correct number of objects in the operation response' do
       response = described_class.call(school_params:)
-      expect(response[:error]).to match(/Error creating school/)
+      expect(response[:error].count).to eq(6)
+    end
+
+    it 'returns the correct type of object in the operation response' do
+      response = described_class.call(school_params:)
+      expect(response[:error].first).to be_a(ActiveModel::Error)
     end
 
     it 'sent the exception to Sentry' do
--- a/lib/concepts/school/operations/create.rb
+++ b/lib/concepts/school/operations/create.rb
@@ -10,8 +10,7 @@ class School
         response
       rescue StandardError => e
         Sentry.capture_exception(e)
-        errors = response[:school].errors.full_messages.join(',')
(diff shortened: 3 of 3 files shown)


sra405:
@cla-bot check

app/controllers/api/schools_controller.rb
```diff
@@ -15,13 +15,14 @@ def show
     end
 
     def create
-      result = School::Create.call(school_params:)
+      @school = School.new(school_params)
```
danhalson, line 18: This isn't wrong, but by calling `.new` directly you're circumventing the library at `lib/concepts/school/operations/create.rb`, I think unless we're making a more generalised move towards this approach we should try and follow the existing pattern. 

I had also spotted that the library was eating errors, so surfacing them here will be good all round.

floehopper: > I think unless we're making a more generalised move towards this approach we should try and follow the existing pattern.

I agree 👍

@tuzz On the off-chance you're seeing this - had you thought about whether/how validation errors would be surfaced from the API?

conorriches: Oh good spot, I'll update the library to provide errored fields which should achieve the same effect

sra405: Yep you did 😄 - https://github.com/RaspberryPiFoundation/editor-api/issues/244

danhalson (approved):
Looks great :)

app/controllers/api/schools_controller.rb
floehopper: @tuzz 
> I copied the existing pattern with the plan to improve it later 

I see 👍

spec/concepts/school/create_spec.rb
```diff
@@ -57,7 +57,7 @@
 
     it 'returns the error message in the operation response' do
       response = described_class.call(school_params:)
-      expect(response[:error]).to match(/Error creating school/)
+      expect(response[:error].count).to eq(6)
```
danhalson, line 60: could we just add another step that checks the response contains at least one sensible thing, like a validation?

conorriches: Sure, done in [bd13bf4](https://github.com/RaspberryPiFoundation/editor-api/pull/270/commits/bd13bf4d52d8c60890cb4e875bac91133cb7a263)

danhalson (approved):
perfect!

chrislo:
@conorriches have we thought about the i18n story around these error messages? I notice we do try to translate/localise strings that appear in editor-standalone, but I'm not sure how that would work here. I can think of two approaches off the top of my head 1) pass a language parameter in API requests so the API can return the correct version of the strings, or 2) return an identifier for each error instead of a string and use that to look up the localisation on the client side. 

I'm also slightly nervous that we're relying on the keys (e.g. `website`) staying in sync between client and server. Have you thought about how to make sure they don't drift apart?

chrislo:
@conorriches - I should say I don't think these queries are caused by things you've introduced in this PR, I think we had the same two issues with the previous approach. I guess I'm mostly interested in having a cross-team approach to them to make sure we keep things consistent.

conorriches:
Hey @chrislo - yeah indeed a good thought, my running assumption was to keep the i18n on the frontend side of things, as we already have i18n set up there, and I was mostly relying on the backend just providing a list of fields that need flagging. 

But it's a good point and I think we should look at i18n more broadly, specifically about how we want the api to respond to locales. The error messages appear to be coming out of rails itself, so passing in a locale and having i18n translations would probably be the best approach if we do want the api to be aware of i18n.

I've [asked in slack](https://raspberrypifoundation.slack.com/archives/C02JBAA2NFP/p1713954442563339) about whether we will need locales - but so far in the frontend we only have the `en` locale, though if we plan to have full i18n support it would be good to define the boundaries for which side the translations will live on.

I'd like to understand more the concerns about keys staying in sync - the reason I felt safe doing this is that if they keys do go out of sync, then the endpoint won't be able to process the form properly as fields will be missing, meaning the form won't submit even without validation messages.

conorriches:
Discussed on Slack and we have captured the requirement in https://github.com/RaspberryPiFoundation/editor-api/issues/272

PR #271, opened 2024-04-23
Fix order of `schools.reference` in `schema.rb`

floehopper:
In #267, this column was removed and then re-added. However, [the squashed merge commit][1] only included a single migration that did not add/remove the column was included, i.e. there was no net effect on `schools.reference`, but the `reference` column was still moved from near the top to near the bottom of the table in `schema.rb`.

Ideally `schema.rb` should reflect the state of the production database and so I've moved `reference` column back to where it was. This also means that other developers won't see unexpected differences in their `schema.rb` when they run the migrations locally.

[1]: https://github.com/RaspberryPiFoundation/editor-api/commit/f3248986273f62017a8d0bef58bdabc6282c6d78

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_04_19_151721) do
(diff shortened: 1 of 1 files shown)


danhalson (approved):
Good spot, thanks

PR #275, opened 2024-04-26
Consistently respond to errors in all environments

chrisroos:
The check for `consider_all_requests_local` (true in development, false in test and production) meant that we were responding with a 500 for these 4 exceptions in development, and with 400, 404, 403 and 422 respectively in test and production. This made is tricky to develop clients of this API because they'd get different responses in development than they would in production.

The commits below have all touched this block of code but none of them explain why the behaviour is as it is, so I'm going to assume it's OK to change as I have here:

- 632f55d5
- c93169b9
- d597bd93
- 86414675


floehopper (approved):
This looks good to me. I believe it may address https://github.com/RaspberryPiFoundation/editor-api/issues/244#issue-2151176700. If so, it might be worth mentioning that in the commit note / PR description or at least updating that issue.

sra405:
@cla-bot /check

sra405:
@cla-bot check

sra405 (approved):
👍

PR #277, opened 2024-04-30
Show email on school, limit actions

danhalson:
Show email on school dashboard
Limit school actions, so only editing allowed on admnistrate


sra405 (approved):
👍 lgtm

PR #278, opened 2024-05-01
Dev-container-tweaks

danhalson:
- adds output at start of test run to show current env_type and database (to aid debugging)
- dev-containers now use their own image, dev specific packages have been removed from the api container
- dev-containers now use their own docker-compose to manage containers
- add a launch.json to debug using ruby LSP


sra405 (approved):
😍

Issue #273, opened 2024-04-25
Separate organisation_id and primary id in Schools table


danhalson:
The Schools model currently expects the `organisation_id` from Profile to be the primary ID. This is problematic as we want to avoid creating an organisation in Profile until the school has been verified, however, the current implementation in the API does this, with a call to `ProfileApiClient.create_organisation`, though the response is stubbed.

In order to allow multiple schools to be created this call has been removed as of #267 and an `organisation_id` added to the Schools model.

This issue is refactor the API to ensure that all necessary calls reference the new `organisation_id`. This will undoubtedly cover all `ProfileApiClient` calls, however after some prodding about it was clear that this alone wasn't enough, and care needs to be taken to ensure the existing relationships aren't broken. Users (creators, teachers and students) will likely need updating too. Updating the factory with new ids should help identify potential problems.

chrisroos:
We removed `School#organisation_id` in PR #281.

PR #281, opened 2024-05-07
Editor API should be canonical source of Organisations

chrisroos:
GitHub issue #273 

Following a discussion with @DanielBrierton, @sra405, @grega, @floehopper, @chrislo and me, we've agreed not to store Schools/Organisations in Profile for now. Editor API is going to be the canonical source of Schools in the short term and we'll only look to store them in Profile if a user need requires us to.

This change allows us to use Editor Standalone to create multiple Schools in this API, because we're no longer assigning the [same ID to every School created](https://github.com/RaspberryPiFoundation/editor-api/blob/6b5575e43dc3b4b689c319d0952eb9140db15418/lib/profile_api_client.rb#L19).

The individual commit notes describe the changes in more detail.

### DIFF
--- a/app/controllers/api/schools_controller.rb
+++ b/app/controllers/api/schools_controller.rb
@@ -15,7 +15,7 @@ module Api
     end
 
     def create
-      result = School::Create.call(school_params:, user_id: current_user.id, token: current_user&.token)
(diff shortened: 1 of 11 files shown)


sra405 (approved):
👍 works like a charm

PR #282, opened 2024-05-08
WIP: Model roles in editor-api instead of profile

chrislo:
We're currently using fake data in editor-api to fill in organisation/role data that we assume will be available from the profile API later. We've decided instead to model organisations (i.e. schools) solely in the editor-api for now (see #281) which means that we now need to model roles too. 

@chrisroos - I haven't started implementing anything here yet, I mostly wanted to see how the current system works, and start removing code that stubs/fakes out the data to more closely reflect the current reality and see what breaks. There seems to be really good test coverage around the API so these test failures might be able to guide us in how to replace the faked functionality with a real implementation. 

## TODO/IDEAS

Change the implementation of `User#school_owner?`, `User#school_teacher?` etc to look up this data in a join table. I *think* a User could `have_many` Schools through `role` or something similar where we have a join table that has

| user_id | school_id | role |
| 123 | 456 | school-owner |

and then change the implementation of methods on User to use that instead. We'd need to make sure that the `school_owner` role was set when a School is created/verified. And we might want to replace the current `School#user_id` with that (and/or model the idea of `requesting` a school separately from `owning` a school). 

The existing tests - many of which are failing in this PR could be fixed by changing the factories/stubs to use this new join table I think. 

I'm not yet worked out how to sequence that into commits that keep the tests passing, so this PR just has a bunch of failing tests for now and should be considered disposable!

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -32,6 +32,8 @@ class User
   end
 
   def org_roles(organisation_id:)
+    return [] unless organisations
+
     organisations[organisation_id.to_s]&.to_s&.split(',')&.map(&:strip) || []
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -70,9 +70,6 @@ class User
       info = info.stringify_keys
       args = info.slice(*ATTRIBUTES)
 
-      # TODO: remove once the UserInfoApi returns the 'organisations' key.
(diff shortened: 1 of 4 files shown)


chrisroos:
I've started work on adding the roles table/model in PR #283.

chrisroos:
Closing in favour of #298.

PR #290, opened 2024-05-14
Avoid caching words used in `PhraseIdentifier` in the database

floehopper:
There are less than 1000 rows in the file which means it really doesn't take long to load and parse, so I don't really see the point in storing them in the database. I've also cached them in memory, so that we don't reload them in every iteration of the `10.times` loop.

Previously the app didn't work unless we'd seeded the database which was an extra complication for developers and deployments. This change removes that complication.

My actual motivation for making the change is to make the deployment of a new instance of the app at test-editor-api.raspberrypi.org simpler.

### DIFF
--- a/app/models/word.rb
+++ b/app/models/word.rb
@@ -1,4 +1,0 @@
-# frozen_string_literal: true
-
-class Word < ApplicationRecord
-end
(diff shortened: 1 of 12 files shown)


chrislo (approved):
Nice improvement, thanks @floehopper 👍

PR #292, opened 2024-05-15
Improve naming of hydra stubs

chrislo:
We found the use of `UserProfileMock#stub_hydra_public_api` to be a bit inconsistent and confusing. This PR extracts a few methods with, we hope, more intention-revealing names.

### DIFF
--- a/spec/features/class_member/creating_a_class_member_spec.rb
+++ b/spec/features/class_member/creating_a_class_member_spec.rb
@@ -90,7 +90,7 @@ RSpec.describe 'Creating a class member', type: :request do
   end
 
   it 'responds 403 Forbidden when the user is a school-student' do
-    stub_hydra_public_api(user_index: user_index_by_role('school-student'))
+    authenticate_as_school_student
 
     post("/api/schools/#{school.id}/classes/#{school_class.id}/members", headers:, params:)
     expect(response).to have_http_status(:forbidden)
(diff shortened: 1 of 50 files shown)


floehopper (dismissed):
A few minor/subjective thoughts:
* Give that the the `authenticate_as_*` methods are stubbing a request rather than actually authenticating the user, I found the names a bit confusing. I wonder whether `stub_authentication_for_*` might be slightly clearer...?
* It might be worth making the `user_index` keyword argument for `#stub_hydra_public_api` required so that we avoid setting a non-intention-revealing default value.
* It's a bit difficult to see from the diffs, but I have a feeling there might be some `let` blocks for e.g. `teacher_index` that are no longer used. And is `#user_index_by_role` still used? If they're not then it would be good to remove them in this PR.

floehopper (approved):

chrisroos:
> A few minor/subjective thoughts:
> 
>     * Give that the the `authenticate_as_*` methods are stubbing a request rather than actually authenticating the user, I found the names a bit confusing. I wonder whether `stub_authentication_for_*` might be slightly clearer...?

As discussed we're going to leave these for now but might revisit the naming in future.

>     * It might be worth making the `user_index` keyword argument for `#stub_hydra_public_api` required so that we avoid setting a non-intention-revealing default value.

Done!

>     * It's a bit difficult to see from the diffs, but I have a feeling there might be some `let` blocks for e.g. `teacher_index` that are no longer used. And is `#user_index_by_role` still used? If they're not then it would be good to remove them in this PR.

We're relying on Rubocop to inform us about unused `let` blocks and it hasn't complained so I think we're OK on that front.

`#user_index_by_role` is still used in quite a few other tests.

PR #294, opened 2024-05-15
More explicit stubbing of UserInfo API

chrisroos:
This is another preparatory commit towards modelling users' roles in Editor API.

The next step is to remove hardcoded user IDs from our factories. Having these explicit stubs in place will make that change easier.


floehopper (approved):
Looks like another good step forward in making things more explicit - good work! 👍

chrislo:
Thanks @floehopper 👍

Issue #285, opened 2024-05-10
Model users' roles in editor-api


chrisroos:
- [x] https://github.com/RaspberryPiFoundation/editor-api/pull/292
- [x] https://github.com/RaspberryPiFoundation/editor-api/pull/293
- [x] https://github.com/RaspberryPiFoundation/editor-api/pull/294
- [x] https://github.com/RaspberryPiFoundation/editor-api/pull/296
- [x] https://github.com/RaspberryPiFoundation/editor-api/pull/298
- [x] https://github.com/RaspberryPiFoundation/editor-standalone/pull/152
- [x] https://github.com/RaspberryPiFoundation/editor-standalone/pull/156
- [ ] https://github.com/RaspberryPiFoundation/editor-api/pull/301

---

- [ ] #306
  - If a user has created a school then they shouldn't be able to see the /schools/new onboarding flow again. At present, the `/api/school` endpoint only returns the school that the user has a role in and not those they've created.
  - Add uniqueness constraint on `School#creator_id` to protect us against users creating more than one school
  - Add validation on `School` that prevents a user with a role in a school from creating a new school
- [ ] #308
- [ ] #307

sra405:
Blocking aspects of this issue have been merged and deployed, other aspects have been separated out and added to the board

chrisroos:
PRs/Issues that have been split out (I've removed them from the description so that all items in the description have been checked):

- https://github.com/RaspberryPiFoundation/editor-api/pull/301
- https://github.com/RaspberryPiFoundation/editor-api/issues/308
- https://github.com/RaspberryPiFoundation/editor-api/issues/307
- https://github.com/RaspberryPiFoundation/editor-api/issues/306

PR #296, opened 2024-05-20
Remove hardcoded IDs from factories

chrisroos:
This is another change toward being able to model users' roles in Editor API (see #285). The changes allow us to:

- Stub both the Hydra UserInfo API (used by `User.from_token` to identify the logged in user) and our UserInfo API (used by `User.from_userinfo` to request information about users given their user_id) to return user data representing a user with arbitrary user_id having one of the valid roles in an organisation with arbitrary organisation_id.
- Create any number of objects from our factories without falling foul of validation or database constraints.

The rules in the system ensure that:

- The `ClassMember#student_id` user has to have the "student" role in the School associated with the ClassMember.
- The `Lesson#user_id` user has to have the "owner" or "teacher" role in the School optionally associated with the Lesson.
- The `Lesson#user_id` user has to match the `SchoolClass#teacher_id` user if Lesson is associated with a SchoolClass.
- The optional `Project#user_id` user has to have the "owner", "teacher" or "student" role in the the School optionally associated with the Project.
- The optional `Project#user_id` user has to match the `Lesson#user_id` user if Project is associated with a Lesson.
- The `SchoolClass#teacher_id` user has to have the "teacher" role in the School associated with the SchoolClass.

Prior to this PR and in order to satisfy these constraints:

Our stubbing of the Hydra UserInfo API and our UserInfo API meant that we were restricted to the following 3 users:

- A user with ID "00000000-0000-0000-0000-000000000000" had the "owner" role for a School with ID "12345678-1234-1234-1234-123456789abc"
- A user with ID "11111111-1111-1111-1111-111111111111" had the "teacher" role for a School with ID "12345678-1234-1234-1234-123456789abc"
- A user with ID "22222222-2222-2222-2222-222222222222" has the "student" role for a School with ID "12345678-1234-1234-1234-123456789abc"

Our factories had the following defaults:

- Schools built from the factory would have an ID of "12345678-1234-1234-1234-123456789abc"
- SchoolClasses built from the factory would have a teacher_id of "11111111-1111-1111-1111-111111111111" (i.e. teacher)
- ClassMembers built from the factory would have a student_id of "22222222-2222-2222-2222-222222222222" (i.e. student)
- Lessons built from the factory would have a user_id of "11111111-1111-1111-1111-111111111111" (i.e. teacher)
- Projects built from the factory would have a user_id of "22222222-2222-2222-2222-222222222222" (i.e. student)

_NOTE_. I've noticed some odd behaviour around some of the above validations where we assume the object is valid when there's no matching user in the UserInfo API. For example, if there's no user in the UserInfo API matching `ClassMember#student_id` then we assume that the `ClassMember` object is valid which seems dangerous to me. I intend to investigate this separately.

Future work (captured in #285):

- Remove hardcoded user_ids and org_ids from users.json
- If Hydra is stubbed for a user_id/school_id then that same stub should be added to the UserInfo API as they're fed from the same data.
- Various validations currently return true if the user is missing from Hydra. This seems wrong.


spec/models/user_spec.rb
```diff
@@ -85,7 +85,7 @@
     end
 
     context 'when no organisations are returned' do
-      let(:ids) { [User::STUDENT_WITHOUT_ORGANISATIONS_ID] }
+      let(:ids) { [SecureRandom.uuid] }
```
floehopper, line 89: Minor: It would be nice to have a one-line explanation in the commit note as to *why* you can now safely do this.

spec/models/lesson_spec.rb
```diff
@@ -4,19 +4,21 @@
 
 RSpec.describe Lesson do
   before do
-    stub_user_info_api_for_teacher(teacher_id: User::TEACHER_ID, school_id: School::ID)
+    stub_user_info_api_for_teacher(teacher_id:, school_id: School::ID)
   end
 
+  let(:teacher_id) { SecureRandom.uuid }
```
floehopper, line 10: Minor: In a few places like this you've added calls to `let` *after* calls to `before`. I know there's nothing technically wrong with this, but I don't think it's as idiomatic/readable as having them the other way around.

lib/hydra_public_api_client.rb
```diff
@@ -15,7 +14,7 @@ def fetch_oauth_user(...)
   def fetch_oauth_user(token:)
     if bypass_auth?
       users = stubbed_data['users']
-      user = users.detect { |attr| attr['id'] == BYPASS_AUTH_USER_ID }
+      user = users.detect { |attr| attr['id'] == '00000000-0000-0000-0000-000000000000' }
```
floehopper, line 17: Minor: I wonder whether there's _some_ value in retaining the `BYPASS_AUTH_USER_ID` constant since it does make it a bit clearer what it's purpose is...

floehopper (approved):
This looks good to me - great work! 🥇 

I've added a few minor comments inline, but none of them should prevent you merging this as is.

Marking as approved!

spec/models/user_spec.rb
floehopper: The same applies to the other commits where you've removed the other similar constants, but I haven't added a comment for each.

spec/features/lesson/creating_a_lesson_spec.rb
```diff
@@ -154,6 +154,7 @@
     it 'responds 403 Forbidden when the user is a school-owner for a different school' do
       school = create(:school, id: SecureRandom.uuid)
       school_class.update!(school_id: school.id)
+      params[:lesson][:school_id] = school.id
```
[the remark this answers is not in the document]
chrisroos: Could spot. I just had to remind myself too...

It's there to satisfy `LessonsController#verify_school_class_belongs_to_school`. Without this change `LessonsController#school` returns the school created in the `let` in the context, and not the school created in this test, which results in the test failing.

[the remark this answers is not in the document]
chrisroos: I've updated the relevant commit note with an explanation.

lib/hydra_public_api_client.rb
chrisroos: I'm planning to do some more work around this so I'm going to leave it as-is for now.

spec/models/user_spec.rb
chrisroos: I've added the following to each of the relevant commits:

> Now that we can stub Hydra's UserInfo API and our UserInfo API to respond with users containing arbitrary user IDs we're no longer restricted to using these constants.

PR #298, opened 2024-05-23
Model users' roles

chrisroos:
GitHub Issue: https://github.com/RaspberryPiFoundation/editor-api/issues/285

The new `Role` model is used to record the roles a `User` has in a given `School`. Valid roles are "owner", "teacher" and "student". Users can have both "owner" and "teacher" roles at the same time, but cannot have either of those roles as well as the "student" role.

A user that creates a school is automatically given the "owner" role when that school is verified.

The frontend designs assume a user is only ever associated with a single school so we add the `Role#users_can_only_have_roles_in_one_school` validation to attempt to enforce that in the data.

The new /api/school endpoint returns the `School` that the logged in user has a role in. We'll need to update [the `getUserSchool` function](https://github.com/RaspberryPiFoundation/editor-standalone/blob/d069a92f3b3eefbbc23a6b8c11593a25a76a5044/src/utils/apiCallHandler.js#L130) to request this endpoint instead of /api/schools.

---

You can manually add roles for the two seed users with:

```
# Add Roles for john.doe@example.com

rails r 'Role.student.create!(user_id: "bbb9b8fd-f357-4238-983d-6f87b99bdbb2", school: School.last)'
rails r 'Role.teacher.create!(user_id: "bbb9b8fd-f357-4238-983d-6f87b99bdbb2", school: School.last)'
rails r 'Role.owner.create!(user_id: "bbb9b8fd-f357-4238-983d-6f87b99bdbb2", school: School.last)'

# Add Roles for jane.doe@example.com

rails r 'Role.student.create!(user_id: "583ba872-b16e-46e1-9f7d-df89d267550d", school: School.last)'
rails r 'Role.teacher.create!(user_id: "583ba872-b16e-46e1-9f7d-df89d267550d", school: School.last)'
rails r 'Role.owner.create!(user_id: "583ba872-b16e-46e1-9f7d-df89d267550d", school: School.last)'
```

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,6 +27,10 @@ class User
     ATTRIBUTES.index_with { |_k| nil }
   end
 
+  def schools
+    School.joins(:roles).merge(Role.where(user_id: id)).distinct
+  end
+
   def organisation_ids
     organisations&.keys || []
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,7 +36,7 @@ class User
   end
 
   def school_roles(school)
-    organisations[school.id.to_s]&.to_s&.split(',')&.map(&:strip) || []
+    Role.where(school:, user_id: id).map(&:role)
   end
 
   def org_role?(organisation_id:, role:)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -47,8 +47,8 @@ class User
     Role.owner.find_by(school_id: organisation_id, user_id: id)
   end
 
-  def school_teacher?(organisation_id:)
-    Role.teacher.find_by(school_id: organisation_id, user_id: id)
+  def school_teacher?(school)
+    Role.teacher.find_by(school:, user_id: id)
   end
 
   def school_student?(school)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -44,7 +44,7 @@ class User
   end
 
   def school_owner?(organisation_id:)
-    org_role?(organisation_id:, role: 'school-owner')
+    Role.owner.find_by(school_id: organisation_id, user_id: id)
   end
 
   def school_teacher?(organisation_id:)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,10 +39,6 @@ class User
     Role.where(school:, user_id: id).map(&:role)
   end
 
-  def org_role?(organisation_id:, role:)
-    org_roles(organisation_id:).include?(role.to_s)
-  end
-
   def school_owner?(school)
     Role.owner.find_by(school:, user_id: id)
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -70,9 +70,6 @@ class User
       info = info.stringify_keys
       args = info.slice(*ATTRIBUTES)
 
-      # TODO: remove once the UserInfoApi returns the 'organisations' key.
-      temporarily_add_organisations_until_the_profile_app_is_updated(args)
-
       new(args)
     end
   end
@@ -89,9 +86,6 @@ class User
     args = auth.extra.raw_info.to_h.slice(*ATTRIBUTES)
     args['id'] = auth.uid
 
-    # TODO: remove once the OmniAuth info returns the 'organisations' key.
-    temporarily_add_organisations_until_the_profile_app_is_updated(args)
-
     new(args)
   end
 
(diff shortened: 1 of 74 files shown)


spec/models/user_spec.rb
```diff
@@ -14,43 +14,6 @@
...
-      end
-
-      context 'with weird extra whitespace in role' do
-        let(:organisations) { { organisation_id => " #{role} " } }
-
-        it { is_expected.to be_truthy }
-      end
-    end
-  end
-
```
floehopper: Minor: It might be worth mentioning in the commit note that most/all of these are unnecessary now the roles are modelled in an enum rather than as a comma-separated string...?

app/models/user.rb
```diff
@@ -70,9 +70,6 @@ def self.from_userinfo(ids:)
       info = info.stringify_keys
       args = info.slice(*ATTRIBUTES)
 
-      # TODO: remove once the UserInfoApi returns the 'organisations' key.
-      temporarily_add_organisations_until_the_profile_app_is_updated(args)
-
```
floehopper, line 75: 🎉

app/models/user.rb
```diff
@@ -89,9 +86,6 @@ def self.from_auth(auth)
     args = auth.extra.raw_info.to_h.slice(*ATTRIBUTES)
     args['id'] = auth.uid
 
-    # TODO: remove once the OmniAuth info returns the 'organisations' key.
-    temporarily_add_organisations_until_the_profile_app_is_updated(args)
-
```
floehopper, line 94: 🎉

app/models/user.rb
```diff
@@ -107,16 +101,6 @@ def self.from_token(token:)
     args['id'] ||= auth['sub']
     args['token'] = token
 
-    # TODO: remove once the HydraPublicApi returns the 'organisations' key.
-    temporarily_add_organisations_until_the_profile_app_is_updated(args)
-
```
floehopper, line 112: 🎉

app/models/user.rb
```diff
@@ -107,16 +101,6 @@ def self.from_token(token:)
...
     new(args)
   end
-
-  def self.temporarily_add_organisations_until_the_profile_app_is_updated(hash)
-    return hash if hash.key?('organisations')
-
-    # Use the same organisation ID as the one from users.json for now.
-    hash.merge!('organisations' => { '12345678-1234-1234-1234-123456789abc' => hash['roles'] })
-  end
```
floehopper, line 121: 🎉 🎉 🎉

spec/features/my_school/showing_my_school_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  it "includes the school details and user's roles in the JSON" do
+    school_json = school.to_json(only: %i[id name website reference address_line_1 address_line_2 municipality administrative_area postal_code country_code verified_at created_at updated_at])
+    expected_data = JSON.parse(school_json, symbolize_names: true).merge(roles: ['owner'])
+
+    get('/api/school', headers:)
+    data = JSON.parse(response.body, symbolize_names: true)
+
+    expect(data).to eq(expected_data)
+  end
```
floehopper, line 27: Minor: There seems to be some duplication in the assertion in this example vs the one below, i.e. both are checking the `roles` are included.

floehopper (dismissed):
This all looks good to me and a *lot* of painstaking work. As well as implementing the new functionality - the code looks so much better - well done! 👏 

I've added a few minor comments inline, but none of them should prevent you from merging.

I share the reservation you mentioned in your commit notes about creating the roles within the stubbing methods, but I understand why you took that approach. It would be nice to sort that out, but I don't think it should block this PR..

Also the fact that you had to delete roles in various specs highlighted the shortcomings of those specs where setup data is being updated for particular contexts or examples. It ought to be possible to use `let` blocks in different contexts to avoid having to do that. However, it was a pre-existing problem and I'm happy that we tackle it outside the scope of this PR.

Marking as approved! ✅

spec/features/my_school/showing_my_school_spec.rb
chrisroos: Good spot. I had intended to remove the second example after making the first one more comprehensive but then forgot! I've removed the second example now.

spec/models/role_spec.rb
```diff
@@ -38,5 +38,107 @@
...
+      let(:user) { build(:user) }
+      let(:school) { build(:school) }
+
+      before do
+        create(:teacher_role, user_id: user.id, school:)
+      end
+
+      it 'allows an owner role to be created for the user and school' do
+        create(:owner_role, user_id: user.id, school:)
+        expect(described_class.where(user_id: user.id, school:).count).to eq(2)
```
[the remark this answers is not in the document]
chrisroos: Done :+1:

spec/models/user_spec.rb
chrisroos: Done :+1:

chrisroos:
Thanks, @floehopper! I've addressed your feedback so would appreciate you re-reviewing when you have a mo.

floehopper (approved):

PR #301, opened 2024-05-29
Require users to exist in model validations

chrisroos:
Tighten up validations of:

- `ClassMember#student_id`
- `Lesson#user_id`
- `Project#user_id`
- `SchoolClass#teacher_id`

Prior to this PR, if the users identified by the IDs in the methods above didn't exist in Hydra then we'd assume the object was valid. We now default to assuming that the object is invalid if the user doesn't exist in Hydra. This feels like a safer default.

### DIFF
--- a/app/models/class_member.rb
+++ b/app/models/class_member.rb
@@ -30,7 +30,7 @@ class ClassMember < ApplicationRecord
     return unless student_id_changed? && errors.blank?
 
     _, user = with_student
-    return if !user || user.school_student?(school)
+    return if user&.school_student?(school)
 
     msg = "'#{student_id}' does not have the 'school-student' role for organisation '#{school.id}'"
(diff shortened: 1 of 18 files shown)


floehopper (changes requested):
As we discussed on a call:

* I found some of the changes in validation logic hard to follow. We agreed it might make things clearer to do some preparatory refactoring e.g. to avoid some of the use of `unless` in combination with `!`.

* There are also some examples where I didn't understand why they'd had to change or even why they'd been like they were in the first place.

* Also I realise this isn't something you've changed, but I feel a bit nervous having model validations that make HTTP requests.

chrisroos:
Thanks, @floehopper. Having discussed this further I'm going to close this and explore the idea of removing Hydra as a dependency in these validations.

PR #310, opened 2024-06-04
Avoid hydra in model validations

chrisroos:
We no longer make a request to the User Info API when validating `ClassMember`, `Lesson`, `Project` or `SchoolClass`. We were previously making requests to the User Info API based on the assumption that users' roles would be stored in Profile. Now that we're storing roles in this app we no longer need to make these requests.

### DIFF
--- a/spec/models/class_member_spec.rb
+++ b/spec/models/class_member_spec.rb
@@ -72,9 +72,9 @@ RSpec.describe ClassMember do
     end
 
     it 'ignores members where no profile account exists' do
-      student_id = SecureRandom.uuid
-      stub_user_info_api_for_unknown_users(user_id: student_id)
-      create(:class_member, student_id:, school_class:)
+      student = create(:student, school:)
+      stub_user_info_api_for_unknown_users(user_id: student.id)
+      create(:class_member, student_id: student.id, school_class:)
 
       student = described_class.all.students.first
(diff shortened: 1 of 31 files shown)


spec/models/class_member_spec.rb
```diff
@@ -4,7 +4,6 @@
 
 RSpec.describe ClassMember do
   before do
-    stub_user_info_api_for(teacher)
     stub_user_info_api_for(student)
```
loiswells97, line 7: I noticed there are still a few of these scattered around and wondered why they are still needed?

chrisroos: There are still other methods that call the User Info API, notably:

- `ClassMember`
  - `.students`
  - `.with_students`
  - `#with_student`
- `Lesson`
  - `.users`
  - `.with_users`
  - `#with_user`
- `Project`
  - `.users`
  - `.with_users`
  - `#with_user`
- `SchoolClass`
  - `.teachers`
  - `.with_teachers`
  - `#with_teacher`

So anywhere that we're calling these methods will require us to stub the User Info API. Ideally I'd prefer the stubbing to be in the example that requires it (rather than in a `before` block which sets it irrespective of whether the example needs it or not) but I felt that was a change too far for this PR.

Does that make sense?

PR #311, opened 2024-06-04
Improve stubbing of hydra

chrisroos:
This PR simplifies the `UserProfileMock` so that we end up with two main methods: `stub_user_info_api_for` and `authenticated_in_hydra_as`. Both of these take a `User` instance and stub the relevant API to return the user's attributes instead of returning hardcoded data from the users.json fixture. It also introduces 3 new `User` factories (`student`, `teacher` and `owner`) which are essentially syntactic sugar around creating the equivalent `Role`s.

I took lots of small steps to get to the final result so there are lots of commits, but each one should hopefully be fairly easy to review in isolation :crossed_fingers:


chrislo (approved):
This makes things much easier to work with and understand, thanks @chrisroos!

PR #312, opened 2024-06-05
Send invitation email when teacher created

chrislo:
[EPIC](https://github.com/orgs/RaspberryPiFoundation/projects/51/views/11?pane=issue&itemId=67112446)

This PR adds functionality to send an email to a teacher when they are invited to join a school. For now the email only has a plain-text version and it includes a link to a page in `editor-standalone` that doesn't currently exist: 

```
You have been invited to join:

Elmwood Primary School

Being part of this school account will allow you to access the school dashboard in the Code Editor and create coding lessons for students.

Join school:

https://editor.raspberrypi.org/en/invitations/<token>

--
Raspberry Pi Foundation
Copyright Raspberry Pi Foundation UK registered charity 1129409
All rights reserved
```

For this reason we're planning to configure both the production and staging deployments of `editor-api` to use the postmark `editor-api-staging` server (via the `POSTMARK_API_TOKEN` environment variable introduced in this PR). This server runs in "sandbox" mode - we can see that emails have been sent, but they are not actually delivered. When the functionality to accept an invitation exists in `editor-standalone` we can reconfigure production to send real emails.


app/models/invitation.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class Invitation < ApplicationRecord
+  belongs_to :school
+  validates :email_address,
+            format: { with: EmailValidator.regexp, message: I18n.t('validations.invitation.email_address') }
+  validate :school_is_verified
+  after_create :send_invitation_email
```
floehopper, line 8: Minor: It _might_ be worth considering using `after_create_commit` vs `after_create` to ensure that we only ever queue an email for sending if the invitation record is successfully committed to the database...?

app/models/invitation.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  generates_token_for :teacher_invitation, expires_in: 30.days do
+    email_address
```
floehopper, line 11: Minor: It might be nice to add a unit test that demonstrates what the call to `Invitation#email_address` in the block is meant to achieve. As I understand it, it should mean the token becomes invalid if it's verified against an invitation record with a different `email_address` than when the token was created.

spec/models/invitation_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    school = build(:school, verified_at: nil)
+    invitation = build(:invitation, school:)
+
+    expect(invitation).not_to be_valid
+  end
+
+  it 'sends an invitation email after create' do
+    school = create(:school, verified_at: Time.zone.now)
+
+    invitation = described_class.create(email_address: 'teacher@example.com', school:)
```
floehopper, line 31: Minor: I might be good to call the bang version of `Invitation.create!` here so we fail fast if the invitation is invalid.

floehopper (approved):
I've added a few minor comments/suggestions, but none of them should prevent you from merging. Approved!

app/models/invitation.rb
chrislo: Good idea, I've made that change in [3800844](https://github.com/RaspberryPiFoundation/editor-api/pull/312/commits/3800844f78d097be81f3c5bc4b0bc2bfc281954c)

chrislo:
Thanks for the review @floehopper! I've addressed your feedback and will fixup those commits before merging. 

I've asked @grega to help set the two environment variables `POSTMARK_API_TOKEN` and `EDITOR_PUBLIC_URL` we introduced in this PR in the relevant deployment environments, so I'll wait until that happens before I merge this so that it's safe for it to be deployed.

floehopper (approved):

PR #324, opened 2024-06-11
Upgrade to rails 7.1

chrislo:
This PR upgrades Rails from 7.0.4.2 to 7.1.3.4. I've primarily done that to be able to use the `generates_token_for` method that was added in Rails 7.1 to generate invitation codes for teachers as part of the [inviting teachers EPIC](https://github.com/orgs/RaspberryPiFoundation/projects/51/views/11?pane=issue&itemId=62107497), but it's a good idea to keep up-to-date with framework changes as the further behind we get the harder it becomes to upgrade.  

To keep the changes to a minimum, I've tried to be conservative in the gems I upgrade, but updates were needed to rails, puma, bullet and rspec-mocks - those changes happen in the first few commits. 

After that I go through each of the new default options introduced in Rails 7.1 one-by-one and explain why it's safe to adopt the default (and make any changes where necessary, although these are minimal). 

This upgrade has introduced a `DEPRECATION WARNING: TestFixtures.fixture_path= is deprecated and will be removed in Rails 7.2` warning when running the tests that I'm not happy about. I haven't been able to track down why that is happening and silence it yet. I didn't want that to hold up this upgrade and subsequent work though, so I've decided not to tackle it in this PR.


bin/rails
```diff
@@ -1,6 +1,4 @@
 #!/usr/bin/env ruby
-# frozen_string_literal: true
-
-APP_PATH = File.expand_path('../config/application', __dir__)
-require_relative '../config/boot'
-require 'rails/commands'
+APP_PATH = File.expand_path("../config/application", __dir__)
+require_relative "../config/boot"
+require "rails/commands"
```
chrisroos, line 4: Do you understand why Rubocop has changed this now? In main this file is using single quotes.

db/schema.rb
```diff
@@ -123,6 +123,14 @@
     t.index ["scheduled_at"], name: "index_good_jobs_on_scheduled_at", where: "(finished_at IS NULL)"
   end
 
+  create_table "invitations", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
chrisroos, line 126: I don't think this table belongs in this commit.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -85,7 +85,7 @@
 # serializer. Therefore, this setting should only be enabled after all replicas
 # have been successfully upgraded to Rails 7.1.
 #++
-# Rails.application.config.active_job.use_big_decimal_serializer = true
+Rails.application.config.active_job.use_big_decimal_serializer = true
```
chrisroos: We do appear to have jobs in `app/jobs` so I wonder whether this requires more investigation to be confident that it's continuing to work as expected?

config/application.rb
```diff
@@ -24,7 +24,7 @@
 module App
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 7.0
+    config.load_defaults 7.1
 
     config.active_support.cache_format_version = 7.1
```
chrisroos, line 29: Can you also remove this line in this commit? Presumably `load_defaults 7.1` includes setting `cache_format_version = 7.1`.

Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/* /var/lib/apt/archives/*.deb
 COPY Gemfile Gemfile.lock /app/
 RUN bundle install --jobs 4 \
-  && bundle binstubs --all --path /usr/local/bundle/bin
+  && bundle binstubs --all --path /usr/local/bundle/bin && bundle binstubs bundler --force
```
chrisroos, line 26: I think it might be clearer if this new command was split onto a new line.

bin/rails
chrisroos: Or maybe Rubocop hasn't touched it, and this is what's been brought in by `rails app:update`. I think it'd be good to switch from double to single quotes in this commit so that the diff is easier to parse. What do you think?

config/application.rb
```diff
@@ -29,6 +29,7 @@ class Application < Rails::Application
     config.autoload_paths << "#{root}/lib"
```
chrisroos, line 33: My reading of https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#config-autoload-lib-and-config-autoload-lib-once is that we can remove this (and maybe the next line) because we're now using `autoload_lib` instead.

db/schema.rb
chrislo: Addressed in 615e487

config/environments/production.rb
```diff
@@ -39,21 +44,31 @@
                     "https://#{ENV.fetch('HEROKU_APP_NAME', '')}.herokuapp.com")
   }
 
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  # config.assume_ssl = true
+
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
chrisroos, line 52: I wonder whether there's a reason this wasn't previously set? It might be worth leaving it commented out.

config/initializers/new_framework_defaults_7_1.rb
chrislo: You're right @chrisroos - there is a single job. I've updated the commit message to read: 

> Accept the default for this option. We only have a single job
UploadJob which takes a hash argument so I don't think we need to
worry about the comment here which suggests we perform a rolling
upgrade to avoid issues with serialized BigDecimal arguments.

config/application.rb
chrislo: Fixed in bb505e4

config/environments/test.rb
```diff
@@ -25,12 +26,12 @@
   }
 
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = false
+  config.consider_all_requests_local = true
```
chrisroos, line 29: I can see this was explicitly set to `false` in https://github.com/RaspberryPiFoundation/editor-api/commit/632f55d526a69d859692489be21ac6b8aa6cc5ed#diff-77f322f5ab0c54b1f2793b339574333dc55889645d910a93ede0fd7aa13b217a but there's no clue as to why. I wonder whether it's worth switching this back to `false`?

bin/rails
chrislo: I'm afraid I don't. In general I'm happy to do whatever makes rubocop happy, so I won't look into it much further unless you're worried it's hiding something?

Dockerfile
chrislo: Fixed in 6e3b5d7

config/application.rb
chrislo: Addressed in 36d8a4d. The next line does appear to be needed.

bin/rails
chrisroos: I had a play. We're seeing this change from single to double quotes because that's how Rails generates these files, and running `rubocop -A` doesn't fix them automatically. You can fix them by running rubocop for the specific files, e.g. `rubocop -A bin/rails`. I think it's worth doing to avoid introducing unnecessary noise in this commit.

chrislo: Thanks for investigating @chrisroos! I've fixed those in ea3d2f9

config/environments/production.rb
chrislo: Addressed in 33c3b1c

config/environments/test.rb
chrislo: Me neither. Given that this is the test environment and I haven't noticed any issues running the tests on this branch, I think I'm going to leave it set to `true`.

chrisroos (approved):
This all looks good to me. Good work, @chrislo :+1:

PR #327, opened 2024-06-11
Add app.json

chrisroos:
Add the Heroku Postgres add-on when the app is deployed to Heroku.

Our Review Apps were failing because our `release` phase was running before Postgres was available. Specifying the Postgres add-on in app.json appears to solve this problem.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,0 +1,10 @@
+{
+  "addons": [
+    {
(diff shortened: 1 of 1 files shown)


chrislo (approved):
Thanks Chris!

PR #329, opened 2024-06-12
Fix fixture_path= deprecation warning

chrislo:
We were seeing

    DEPRECATION WARNING: TestFixtures.fixture_path= is deprecated and
    will be removed in Rails 7.2. Use .fixture_paths= instead.

when running spec. Upgrading rspec-rails to 6.1.1 silenced the deprecation warning that was being raised, but replaced it with a note at the end of the test run

    Rails 7.1 has deprecated the singular fixture_path in favour of an
    array.You should migrate to plural

Since we're not using fixtures in this project the simplest way to fix that was to remove the `config.fixture_path` call from `spec/rails_helper.rb` altogether.


chrisroos (approved):
Thank you!

PR #334, opened 2024-06-17
Create schools in Profile API upon verification

chrisroos:
This PR expands `SchoolVerificationService#verify` so that it uses the Profile API to create the school in Profile during verification. If any part of verification fails (updating the school, creating the teacher and owner roles, or creating the school in Profile API) then we abort verification and display a message saying that verification failed.

- [x] Set `PROFILE_API_KEY` in staging
- [x] Set `PROFILE_API_KEY` in production

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -19,8 +19,14 @@ class School < ApplicationRecord
   validates :creator_agree_terms_and_conditions, presence: true, acceptance: true
   validates :rejected_at, absence: { if: proc { |school| school.verified? } }
   validates :verified_at, absence: { if: proc { |school| school.rejected? } }
+  validates :code,
+            uniqueness: { allow_nil: true },
+            presence: { if: proc { |school| school.verified? } },
+            absence: { unless: proc { |school| school.verified? } },
+            format: { with: /\d\d-\d\d-\d\d/, allow_nil: true }
   validate :verified_at_cannot_be_changed
   validate :rejected_at_cannot_be_changed
+  validate :code_cannot_be_changed
 
   before_validation :normalize_reference
 
@@ -44,7 +50,15 @@ class School < ApplicationRecord
   end
 
   def verify!
-    update!(verified_at: Time.zone.now)
+    attempts = 0
+    begin
+      update!(verified_at: Time.zone.now, code: SchoolCodeGenerator.generate)
+    rescue ActiveRecord::RecordInvalid => e
+      raise unless e.record.errors[:code].include?('has already been taken') && attempts <= 5
+
+      attempts += 1
+      retry
+    end
   end
 
   def reject
@@ -64,5 +78,9 @@ class School < ApplicationRecord
 
   def rejected_at_cannot_be_changed
     errors.add(:rejected_at, 'cannot be changed after rejection') if rejected_at_was.present? && rejected_at_changed?
+  end
+
+  def code_cannot_be_changed
+    errors.add(:code, 'cannot be changed after verification') if code_was.present? && code_changed?
   end
 end
--- a/lib/profile_api_client.rb
+++ b/lib/profile_api_client.rb
@@ -4,6 +4,20 @@ class ProfileApiClient
   class << self
     # TODO: Replace with HTTP requests once the profile API has been built.
 
+    def create_school(token:, school:)
+      response = connection.post('/api/v1/schools') do |request|
+        apply_default_headers(request, token)
+        request.body = {
+          id: school.id,
+          schoolCode: school.code
+        }.to_json
+      end
+
+      raise "School not created in Profile API. HTTP response code: #{response.status}" unless response.status == 201
+
+      JSON.parse(response.body)
+    end
+
     # The API should enforce these constraints:
     # - The token has the school-owner or school-teacher role for the given organisation ID
     # - The token user or given user should not be under 13
@@ -180,6 +194,20 @@ class ProfileApiClient
       # code so that SchoolOwner::Remove propagates the error in the response.
       response = {}
       response.deep_symbolize_keys
+    end
+
+    private
+
+    def connection
+      Faraday.new(ENV.fetch('IDENTITY_URL'))
+    end
+
+    def apply_default_headers(request, token)
+      request.headers['Accept'] = 'application/json'
+      request.headers['Authorization'] = "Bearer #{token}"
+      request.headers['Content-Type'] = 'application/json'
+      request.headers['X-API-KEY'] = ENV.fetch('PROFILE_API_KEY')
+      request
     end
   end
 end
--- a/lib/profile_api_client.rb
+++ b/lib/profile_api_client.rb
@@ -5,6 +5,8 @@ class ProfileApiClient
     # TODO: Replace with HTTP requests once the profile API has been built.
 
     def create_school(token:, school:)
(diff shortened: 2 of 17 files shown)


spec/lib/school_code_generator_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      allow(Random).to receive(:new).and_return(random)
+
+      expect(described_class.generate).to eq('00-01-23')
+    end
+
+    it 'generates a string containing 3 pairs of digits' do
+      expect(described_class.generate).to match(/\d\d-\d\d-\d\d/)
+    end
+
+    it 'generates a different code each time' do
```
chrislo, line 19: I think there's a small chance that this algorithm could generate the same code twice in a row, which makes this test potentially flakey. I wonder if `generate` could take an optional seed to pass to [Random.new](https://docs.ruby-lang.org/en/master/Random.html) so that the test could be made deterministic?

lib/school_code_generator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class SchoolCodeGenerator
+  MAX_CODE = 1_000_000
+
+  def self.generate
+    number = Random.new.rand(MAX_CODE)
```
floehopper, line 7: Minor: I think you could just use `Kernel#rand` here: https://docs.ruby-lang.org/en/3.2/Kernel.html#method-i-rand

app/models/school.rb
```diff
@@ -44,7 +50,15 @@ def rejected?
...
-    update!(verified_at: Time.zone.now)
+    attempts = 0
+    begin
+      update!(verified_at: Time.zone.now, code: SchoolCodeGenerator.generate)
+    rescue ActiveRecord::RecordInvalid => e
+      raise unless e.record.errors[:code].include?('has already been taken') && attempts <= 5
+
+      attempts += 1
+      retry
+    end
```
floehopper, line 61: Minor: I don't think you need the `begin` / `end` statements here, do you?

lib/school_code_generator.rb
```diff
@@ -0,0 +1,14 @@
...
+    code = format('%06d', number)
+
+    code.match(/(\d\d)(\d\d)(\d\d)/) do |m|
+      [m[1], m[2], m[3]].join('-')
+    end
```
floehopper, line 12: Minor: Is this formatting definitely something we want to store in the database, i.e. is it not just a presentational concern?

lib/profile_api_client.rb
```diff
@@ -5,6 +5,8 @@ class << self
     # TODO: Replace with HTTP requests once the profile API has been built.
 
     def create_school(token:, school:)
+      return { 'id' => school.id, 'schoolCode' => school.code } if ENV['BYPASS_OAUTH'].present?
```
chrislo, line 8: Is the BYPASS_OAUTH variable a development thing, so that you don't have to run profile locally?

chrislo (approved):
Looks good Chris!

lib/profile_api_client.rb
```diff
@@ -181,5 +195,19 @@ def delete_school_student(token:, student_id:, organisation_id:)
       response = {}
       response.deep_symbolize_keys
     end
+
+    private
+
+    def connection
+      Faraday.new(ENV.fetch('PROFILE_API_URL'))
```
floehopper, line 204: Note: We'll need to make sure the two new env vars are set in the relevant environments before deploying, because these calls to `Hash#fetch` will fail fast.

floehopper (approved):
This all looks good to me. I've added a few thoughts inline, but nothing that should prevent you from merging.

spec/lib/school_code_generator_spec.rb
chrislo: Or you could stub `Rand.new` thinking about it - as we don't really need to test that method.

lib/profile_api_client.rb
chrisroos: Yep. It's used in a few places to prevent this app communicating with Profile/Hydra/UserInfo API. Although there aren't any checks that we're also in development (or test) environments which would feel a bit safer.

app/models/school.rb
chrisroos: I started without them but it meant that `attempts` was always reset to 0 after the call to `retry`. I considered setting an instance variable (e.g. `@verification_attempts`) in `after_initialize` but didn't think it bought enough to justify moving the counter away from where it's used.

lib/school_code_generator.rb
chrisroos: I wondered the same thing but settled on storing it in this format as this is what we send to the Profile API when creating the school.

lib/school_code_generator.rb
chrisroos: Hmm. I started out using `Kernel#rand` but switched to `Random#rand` as I mistakenly thought `Kernel#rand` didn't specify whether the generated value was inclusive of exclusive of the `max` argument. Do you think it's worth switching to `Kernel#rand`?

spec/lib/school_code_generator_spec.rb
chrisroos: Good shout :+1:

chrisroos: Hmm, having tried a few things I'm less sure how useful it is.

We could stub `Random#rand` but then this feels like it's duplicating the first test:

```ruby
it 'generates a different code each time' do
  random = instance_double(Random)
  allow(random).to receive(:rand).and_return(123, 456)
  allow(Random).to receive(:new).and_return(random)

  expect(described_class.generate).not_to eq(described_class.generate)
end
```

Or we could stub `Random.new_seed` (which is used to set the `seed` if no explicit seed is passed to `Random.new`) but I don't think that's enough to guarantee that we won't get duplicate numbers, is it?

```ruby
it 'generates a different code each time' do
  allow(Random).to receive(:new_seed).and_return(1, 2)

  expect(described_class.generate).not_to eq(described_class.generate)
end
```

I'm tempted to leave it as-is. What do you think?

chrislo: I think what I was imagining was that if you could set the seed in test, then you'd always get the same sequence of generated numbers: 

```
irb(main):001> seed = 123
=> 123
irb(main):002> rng = Random.new(seed)
=> #<Random:0x0000000110093088>
irb(main):003> rng.rand(100)
=> 66
irb(main):004> rng.rand(100)
=> 92
```

so you could write the test to expect those specific numbers (or expect that they're different to one another as you do now). 

But the chance of generating the same number twice in a row is 1:1000000 * 1:1000000 or 1 in a billion! So I don't think we do really need to worry about it 😂

chrisroos: Ah, I see. So use the same seed to guarantee a different number each time, rather than two different seeds.

However, I'm going to leave it as-is given that the chance of collision seems so small. Thank you!

lib/school_code_generator.rb
floehopper: I don't think it matters much

app/models/school.rb
floehopper: Doh! Of course - I completely missed that 👍

chrisroos:
@sra405's confirmed that the API keys have been set in staging and production so I'm going to merge this now.

PR #335, opened 2024-06-17
291-setup-seed-data

danhalson:
- Add tasks to seed data, these are called in development only via `db:prepare`
- Remove unused argument 'current_user' from verification service
- Fixes auth bypass to use BYPASS_OAUTH everywhere (as used in `rpi_auth`)
- Fixes a dev-container issue that prevented docker control from within the container, and debugging


sra405 (approved):
🙌  that's really nice and well tested!!

PR #338, opened 2024-06-20
Filtering lessons by class

loiswells97:


spec/features/lesson/listing_lessons_spec.rb
```diff
@@ -69,6 +70,22 @@
     expect(data.size).to eq(1)
   end
 
+  it 'does not include lessons from outside the class if school_class_id provided' do
+    get("/api/lessons?school_class_id=#{school_class.id}", headers:)
+    data = JSON.parse(response.body, symbolize_names: true)
```
danhalson, line 75: I'd feel more confident in this step if there was a lesson associated with a school_class that isn't being returned, but it doesn't look like there is?

spec/features/lesson/listing_lessons_spec.rb
```diff
@@ -69,6 +70,22 @@
...
+
+  it 'includes lessons from the class if school_class_id provided' do
+    lesson.update!(school_class_id: school_class.id)
+
+    get("/api/lessons?school_class_id=#{school_class.id}", headers:)
+    data = JSON.parse(response.body, symbolize_names: true)
+
+    expect(data.size).to eq(1)
+  end
+
```
danhalson, line 88: also might be worth adding a step to check that there's no impact on the archived state of a lesson, so checking that if both `include_archived` and `school_class_id` are passed, an archived class is returned (and maybe the inverse?)

spec/features/lesson/listing_lessons_spec.rb
loiswells97: Thanks Dan, added another test for this 👍

spec/features/lesson/listing_lessons_spec.rb
loiswells97: Thanks Dan, have added a couple more test to reflect this ✅

sra405 (approved):
👍 nice changes + tests

PR #340, opened 2024-06-20
Add DatabaseCleaner gem to clear db between task tests

danhalson:
Add DatabaseCleaner gem to clear db between task tests


chrisroos (approved):
I've tested it locally and no longer see the failing test so LGTM :+1:

chrisroos (approved):

PR #342, opened 2024-06-21
Auto-creating project when lesson created

loiswells97:


floehopper (commented):
I started trying to review this, but I'm struggling to understand _why_ some of the changes are being made, because there's no explanation in the commit notes or in the PR description.

For example:
- why is the `Lesson` being changed so that it only has a single `Project` associated with it?
- are we now saying that owners cannot create lessons?

loiswells97:
Hi James, thanks for your comments - here are the answers to your questions:

> * why is the `Lesson` being changed so that it only has a single `Project` associated with it?
I'm pretty sure this is something we agreed?

> * are we now saying that owners cannot create lessons?
I think all we're saying is that owners can't create lessons for other teachers (for now at least) because we don't have any designs for this and since we are forcing all owners to also be teachers, keeping this in place blocks them being able to create lessons for themselves.

spec/features/lesson/updating_a_lesson_spec.rb
```diff
@@ -9,6 +9,7 @@
   end
 
   let(:headers) { { Authorization: UserProfileMock::TOKEN } }
+  # let(:project) { create(:project, user_id: owner.id) }
```
sra405, line 12: 🧟

loiswells97: 🚮

lib/concepts/lesson/operations/create.rb
```diff
@@ -5,15 +5,24 @@ class Create
...
+        response[:error] = "Error creating lesson: #{e}"
         response
       end
+
+      private
+
+      def build_lesson(lesson_hash)
+        new_lesson = Lesson.new(lesson_hash.except(:project_attributes))
+        project_params = lesson_hash[:project_attributes].merge({ user_id: lesson_hash[:user_id],
+                                                                  school_id: lesson_hash[:school_id] })
```
sra405, line 22: 🔧 set `lesson_id` too?

loiswells97: Ah.... yes, I guess so 👀

lib/concepts/lesson/operations/create.rb
```diff
@@ -5,15 +5,25 @@ class Create
...
+
+      private
+
+      def build_lesson(lesson_hash)
+        new_lesson = Lesson.new(lesson_hash.except(:project_attributes))
+        project_params = lesson_hash[:project_attributes].merge({ user_id: lesson_hash[:user_id],
+                                                                  school_id: lesson_hash[:school_id],
+                                                                  lesson_id: new_lesson.id })
+        new_lesson.project = Project.new(project_params)
+        new_lesson
```
sra405, line 25: 🔧 I think it'd be best to do this as a [transaction](https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html) so we know both the lesson and project are created or its rolled back so neither are 👍

lib/concepts/lesson/operations/create.rb
```diff
@@ -5,15 +5,25 @@ class Create
...
         response = OperationResponse.new
-        response[:lesson] = Lesson.new(lesson_params)
+        response[:lesson] = build_lesson(lesson_params)
         response[:lesson].save!
         response
       rescue StandardError => e
         Sentry.capture_exception(e)
-        errors = response[:lesson].errors.full_messages.join(',')
-        response[:error] = "Error creating lesson: #{errors}"
+        response[:error] = "Error creating lesson: #{e}"
```
chrislo, line 13: I think this change might mean that you lose the validation error messages that were previously added to `lesson` if the record couldn't be saved. Without context I'm not sure if we need those on the front-end or not, but perhaps you could merge the validation errors from Lesson and Project and return those?

spec/concepts/lesson/create_spec.rb
```diff
@@ -63,13 +73,57 @@
...
+    end
+
     it 'returns a failed operation response' do
       response = described_class.call(lesson_params:)
       expect(response.failure?).to be(true)
     end
 
     it 'returns the error message in the operation response' do
       response = described_class.call(lesson_params:)
+      pp response[:error]
```
sra405, line 126: 🧟

lib/concepts/lesson/operations/create.rb
loiswells97: Ah, yes, you're right, I will revert - thanks!

sra405 (approved):
👍 nice! Awkward one but it's looking good now 🥳

PR #343, opened 2024-06-24
Remove dependency on FactoryBot, make tasks atomic

danhalson:
- Remove dependency on FactoryBot
- Make tasks atomic


sra405 (approved):
👍 nice

PR #344, opened 2024-06-24
Improve names, and remove special char

danhalson:
Improve names, and remove special char


sra405 (approved):
😅 how'd that get there!!

PR #347, opened 2024-06-25
Fix projects admin

chrisroos:
Installing libvips in the Docker container and replacing `ActiveStorage::Current.host=` with `ActiveStorage::Current.url_options=` means that /admin/projects renders without errors and the images can be displayed.


floehopper (approved):
Looks good to me! 👍

chrisroos:
Thanks @floehopper!

PR #348, opened 2024-06-25
Prepare for generating school codes

chrisroos:
This is mostly a refactoring PR in preparation for creating a school in Profile API upon verification in Editor API. Once we're creating schools in Profile API we'll need to keep them in sync with the data in Editor API. To make that easier this PR restricts a verified school from being rejected so that we don't have to handle updating/deleting schools in Profile API (at least for now).

This PR also:

- tightens up validation so that a school can only have one of `verified_at` or `rejected_at` set at a time, and so that once set those values can't be changed.
- adds tests around verification and rejection of schools in the admin UI before then simplifying those controller actions.
- fixes a bug in `SchoolVerificationService#verify` whereby a failed verification would appear to have been successful to the user.

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -47,6 +47,10 @@ class School < ApplicationRecord
     update!(verified_at: Time.zone.now)
   end
 
+  def reject
+    update(rejected_at: Time.zone.now)
+  end
+
   private
 
   # Ensure the reference is nil, not an empty string
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -31,6 +31,10 @@ class School < ApplicationRecord
     User.from_userinfo(ids: creator_id).first
   end
 
+  def verified?
+    verified_at.present?
+  end
+
   private
 
   # Ensure the reference is nil, not an empty string
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -19,6 +19,7 @@ class School < ApplicationRecord
   validates :creator_agree_terms_and_conditions, presence: true, acceptance: true
   validates :rejected_at, absence: { if: proc { |school| school.verified? } }
   validates :verified_at, absence: { if: proc { |school| school.rejected? } }
+  validate :verified_at_cannot_be_changed
 
   before_validation :normalize_reference
 
@@ -46,5 +47,9 @@ class School < ApplicationRecord
   # Ensure the reference is nil, not an empty string
   def normalize_reference
     self.reference = nil if reference.blank?
+  end
+
+  def verified_at_cannot_be_changed
+    errors.add(:verified_at, 'cannot be changed after verification') if verified_at_was.present? && verified_at_changed?
   end
 end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -20,6 +20,7 @@ class School < ApplicationRecord
   validates :rejected_at, absence: { if: proc { |school| school.verified? } }
   validates :verified_at, absence: { if: proc { |school| school.rejected? } }
   validate :verified_at_cannot_be_changed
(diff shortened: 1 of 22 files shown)


spec/services/school_verification_service_spec.rb
```diff
@@ -10,21 +10,62 @@
...
+      it 'returns true' do
+        expect(service.verify).to be(true)
+      end
     end
 
-    it 'grants the creator the owner role for the school' do
-      expect(school_creator).to be_school_owner(school)
+    describe 'when school cannot be saved' do
+      before do
+        school.update_attribute(:website, 'invalid') # rubocop:disable Rails/SkipsModelValidations
```
floehopper, line 36: Minor: It would be nicer to use a `let` blocks to avoid having to update the school in the database and avoid having to disable the rubocop rule.

spec/features/admin/schools_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+RSpec.describe 'Schools', type: :request do
+  let(:admin_user) { create(:admin_user) }
+
+  before do
+    sign_in_as(admin_user)
+  end
+
+  describe 'GET #index' do
+    it 'responds 200' do
+      get '/admin/schools'
```
floehopper, line 14: Minor: It might be nicer to use `admin_schools_path` here...?

app/models/school.rb
```diff
@@ -47,6 +47,10 @@ def verify!
     update!(verified_at: Time.zone.now)
   end
 
+  def reject
+    update(rejected_at: Time.zone.now)
+  end
+
```
floehopper, line 53: Is there any reason you've made `School#reject` a non-bang method calling `School#update` whereas `School#verify!` is a bang method calling `School#update!`?

floehopper (approved):
I've added a few notes inline, but otherwise this all looks good to me and I don't think any of my notes should prevent you from merging. 👍

app/models/school.rb
chrisroos: I've added the following to the commit note. Do you think this is clear enough?

> To complement `School#verify!`. Note that this doesn't need to be a bang method because the return value of calling `update` is passed through `SchoolVerificationService#reject` and back to `Admin::SchoolsController#reject`. The `School#verify!` method needs to raise on error so that it triggers a transaction rollback in `SchoolVerificationService#verify`.

spec/features/admin/schools_spec.rb
chrisroos: Good spot! I've amended the commit.

spec/services/school_verification_service_spec.rb
chrisroos: I've addressed this in a new commit titled "Avoid update_attribute in SchoolVerificationService spec".

chrisroos:
Thanks @floehopper. Would you mind reviewing/approving it again now that I've addressed your comments, please? I've made the following changes since your last review:

- Used the `admin_schools_path` helper method in the commit titled "Add tests for Admin::SchoolsController"
- Removed the need to disable a RuboCop rule in a new commit titled "Avoid update_attribute in SchoolVerificationService spec"
- Updated the message of commit titled "Extract School#reject" to explain why it's not a bang method.

spec/services/school_verification_service_spec.rb
```diff
@@ -32,13 +38,11 @@
     end
 
     describe 'when school cannot be saved' do
-      before do
-        school.update_attribute(:website, 'invalid') # rubocop:disable Rails/SkipsModelValidations
-      end
+      let(:website) { 'invalid' }
```
floehopper, line 41: Nice! 😄

floehopper (approved):

PR #368, opened 2024-07-02
Encrypt teacher invitation email address

chrislo:
## Status

Related to: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/159

## What's changed?

This PR ensures that the `email_address` field on the `teacher_invitations` table is encrypted. We can't currently avoid storing this PII in the editor-api database (as we need to use the email adress to send the invitation email) but this change helps mitigate the risk of someone getting access to the production database. 

We haven't sent any teacher invitation emails in production yet, so we don't need to worry about migrating or encrypting any existing data. 

## Steps to perform before deploying to production

We need to set the following ENV variables on non-local environments: 

ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY
ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY
ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT

Rails provides a command, `bin/rails db:encryption:init` to generate suitable values for these, but any long non-guessable strings should be fine.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,9 @@ jobs:
       POSTGRES_USER: choco
       POSTGRES_HOST: "127.0.0.1"
       RAILS_ENV: test
+      ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY: primary-key
+      ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY: deterministic-key
+      ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT: derivation-salt
     steps:
       - checkout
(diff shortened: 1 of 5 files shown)


sra405:
I've set the required env vars in staging and prod so will work when merged but see [this thread](https://raspberrypifoundation.slack.com/archives/C02JBAA2NFP/p1719926002285179) because I couldn't get it working in the review app, staging might be different though

sra405 (approved):
👍 lgtm but might need some tweaking depending on how it pans out in staging

sra405 (approved):
👍 in fact here's another approval 😄

PR #369, opened 2024-07-02
Fix copy in teacher invitation email

chrislo:
Replacing "create coding lessons for students" with "create coding projects for students" as per the copy review with Max and Mel.


loiswells97 (approved):
👍

PR #381, opened 2024-07-08
Add POSTMARK_API_TOKEN to .env.example

chrislo:
The POSTMARK_API_TOKEN variable introduced in 4ab58c49c494 doesn't need to be set in development, but it is useful to document how to set it if we want to use Postmark in development (e.g. when developing/testing HTML emails).


loiswells97 (approved):
Looks great, thanks Chris 👍

PR #383, opened 2024-07-11
Update user model to better support students

sra405:
## Status

- Related to https://github.com/RaspberryPiFoundation/editor-standalone/issues/236

## What's changed?

- student IDs are returned from profile with a `student:` prefix (see [here](https://github.com/RaspberryPiFoundation/profile/blob/main/app/lib/subject.js#L23)) - this broke things like the `my_school` endpoint
- add factory for student specific attributes
- separate out student user tests


app/models/user.rb
```diff
@@ -94,7 +94,7 @@ def self.from_token(token:)
     auth = auth.stringify_keys
     args = auth.slice(*ATTRIBUTES)
 
-    args['id'] ||= auth['sub']
+    args['id'] ||= auth['sub'].sub('student:', '')
```
danhalson, line 97: this looks a little hacky, why would an id contain the prefix `student:`?

spec/models/user_spec.rb
```diff
@@ -47,26 +47,77 @@
...
+        expect(user.email).to eq 'school-owner@example.com'
+      end
+
+      it 'returns a user without a username' do
+        expect(user.username).to be_nil
+      end
+
+      context 'when BYPASS_OAUTH is true' do
+        around do |example|
+          ClimateControl.modify(BYPASS_OAUTH: 'true') do
```
danhalson, line 79: I like ClimateControl 👍

spec/models/user_spec.rb
```diff
@@ -47,26 +47,77 @@
...
+      before do
+        authenticated_in_hydra_as(student)
+      end
+
+      it 'returns an instance of the described class' do
+        expect(user).to be_a described_class
+      end
+
+      it 'returns a user with the correct ID' do
+        expect(user.id).to eq student.id
```
danhalson, line 107: it's not obvious to me that the id that's being passed into to the model is starting with a `student:` prefix, assuming we do need the logic in the model to handle this (which I'm sure we do) it would be good to add an explicit test around this, unless I'm missing something implicit!

app/models/user.rb
sra405: it does feel a little hacky but I found student IDs are always prefixed (see [here](https://github.com/RaspberryPiFoundation/profile/blob/main/app/lib/subject.js#L21-L26)) which was breaking everything on our end and thought this was the best way to solve that

spec/models/user_spec.rb
sra405: no you're right, I don't actually think the id here is prefixed!

spec/models/user_spec.rb
```diff
@@ -47,26 +47,77 @@
...
+      before do
+        authenticated_in_hydra_as(student, :student)
+      end
+
+      it 'returns an instance of the described class' do
+        expect(user).to be_a described_class
+      end
+
+      it 'returns a user with the correct ID' do
+        expect(user.id).to eq "student:#{student.id}"
```
danhalson, line 107: I'd expect this to pass as `expect(user.id).to eq student.id` if the `from_token` logic is being called correctly, right?

loiswells97 (approved):
Looks good, let's get it in 👍

PR #391, opened 2024-07-18
List class member students

sra405:
## What's changed?

Updated class members controller to utilise `SchoolStudents:List` and return the student details in a nested `student` attribute in the class members response.

Previous Profile API mocks were for a single student so I expanded this to allow and test for a filtered subset using the `student_ids` array

### DIFF
--- a/spec/concepts/school_student/list_spec.rb
+++ b/spec/concepts/school_student/list_spec.rb
@@ -5,49 +5,99 @@ require 'rails_helper'
 RSpec.describe SchoolStudent::List, type: :unit do
   let(:token) { UserProfileMock::TOKEN }
   let(:school) { create(:school) }
-  let(:student) { create(:student, school:) }
+  let(:students) { create_list(:student, 3, school:) }
 
-  before do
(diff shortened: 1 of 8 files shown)


spec/concepts/school_student/list_spec.rb
```diff
@@ -5,49 +5,99 @@
...
+    it 'returns a filtered school students JSON array' do
+      response = described_class.call(school:, token:, student_ids:)
+      expect(response[:school_students].size).to eq(2)
+    end
+
+    it 'returns the filtered school students in the operation response' do
+      response = described_class.call(school:, token:, student_ids:)
+      filtered_students.each do |student|
+        expected_user = User.new(id: student.id, name: student.name, username: student.username)
+        expect(response[:school_students]).to include(expected_user)
```
loiswells97, line 76: I'm a bit confused about why we're expecting a User object to be returned in the array here? More for my own understanding than anything 🤔

loiswells97: Never mind, I spotted the answer further down 😅 👍

loiswells97 (approved):
Yay, looks great! Thanks so much for sorting this 🎉

PR #392, opened 2024-07-19
Class-member-create-allow-multiple-student-ids

sra405:
## What's changed?

- Additional class_members endpoint for batch creation
- both create and create_batch use the same create concept
- class members create methods first check for the students in profile via `SchoolStudent::List`
- the create concept utilises the `student` instance over ID strings
- the `student` attributes are nested in the `class_members` response
- failed student class member creations are added to an `errors` object which is returned along with the successful creations


spec/concepts/class_member/create_spec.rb
```diff
@@ -5,61 +5,138 @@
...
+        let(:students) { [different_school_student] }
+
+        it 'does not create a class member' do
+          expect { described_class.call(school_class:, students:) }.not_to change(ClassMember, :count)
+        end
+
+        it 'returns a successful operation response' do
+          response = described_class.call(school_class:, students:)
+          expect(response.success?).to be(true)
+        end
```
loiswells97, line 84: Why a successful response here?

loiswells97: Is it so we can surface the errors from each student?

spec/features/class_member/creating_a_batch_of_class_members_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  context 'with invalid params' do
+    let(:invalid_params) { { class_member: { student_ids: ' ' } } }
+
+    before do
+      authenticated_in_hydra_as(owner)
+      stub_profile_api_list_school_students(school:, student_attributes: [])
+    end
+
+    it 'responds 400 Bad Request when params are missing' do
```
loiswells97, line 75: Does this description need updating?

spec/features/class_member/creating_a_class_member_spec.rb
```diff
@@ -23,55 +18,76 @@
...
+    it 'responds with the class member JSON' do
+      post("/api/schools/#{school.id}/classes/#{school_class.id}/members", headers:, params:)
+      data = JSON.parse(response.body, symbolize_names: true)
 
-  it "responds with nil attributes for the student if their user profile doesn't exist" do
-    student_id = SecureRandom.uuid
-    new_params = { class_member: params[:class_member].merge(student_id:) }
+      student_ids = data.pluck(:student_id)
+      expect(student_ids).to eq([student.id])
+    end
```
loiswells97, line 51: Wait, so posting to the single endpoint is still going to return a list (of one)? If so, this isn't what I expected

loiswells97 (commented):
Just a few small questions/comments ☝️

spec/features/class_member/creating_a_batch_of_class_members_spec.rb
sra405: Yes it does 😄 thanks

spec/concepts/class_member/create_spec.rb
sra405: yes its exactly that. So rails will pick up if there's a `response.error` however not `response.errors` so I'm adding to the former for complete request failures but given its a batch endpoint I'm adding to `errors` for the individual cases that error. A little implicit but I think it makes sense and in combination with an empty `class_members` array and `errors` being returned we should be able to surface and expand on this on the frontend.

Let me know what you think though 👍

loiswells97: I think that makes sense, let's go with that, and we can always change it if it surfaces some weird edge cases 👍

spec/features/class_member/creating_a_batch_of_class_members_spec.rb
sra405: done 👍

spec/features/class_member/creating_a_class_member_spec.rb
sra405: and done 👍 had fun with jbuilder partials for this one!

loiswells97 (approved):
Looks great, let's get this out there!

PR #406, opened 2024-08-08
Fix_env_variables_to_allow_both_api_and_dashboard

danhalson:
## Status

- Ready for review

## What's changed?

- Use a different env var locally for the profile api, to allow simultaneous use of administrate and the api (at long last)
  - This is a local only env var since staging and prod share the same url
- Remove network from docker-compose for now, as unused
- Add extensions to devcontainer.json
- Fix smee architecture to amd64

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,20 +7,20 @@ AWS_SECRET_ACCESS_KEY=changeme
 
 GITHUB_WEBHOOK_SECRET=test_token
 
-POSTGRES_DB=development
 POSTGRES_HOST=db
 POSTGRES_PASSWORD=password
 POSTGRES_USER=postgres
 
-USERINFO_API_URL=http://localhost:6000
+USERINFO_API_URL=http://host.docker.internal:6000 # Internal docker address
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -7,20 +7,20 @@ AWS_SECRET_ACCESS_KEY=changeme
 
 GITHUB_WEBHOOK_SECRET=test_token
 
-POSTGRES_DB=development
```
conorriches, line 10: Just checking this is meant to be deleted?

danhalson: yeh, having this set has caused issues recently, since it overwrites the default db name for both development and test, and lands them up sharing a database which is a bit weird and causes rails to behave weirdly

danhalson: the defaults of `choco_cake_development` and `choco_cake_test` are good, and make me think of tasty food which I think is a positive thing, if a little misleading

conorriches: Roger!

conorriches (approved):
SMASHING!

Issue #405, opened 2024-08-08
Explicit CORS Update


sra405:
Currently all our CORS URLs are set in environment variables in terraform for [staging](https://github.com/RaspberryPiFoundation/terraform/blob/5e4884626fc4d9fb84fa8b1083d7aab8d6f690c3/projects/editor-api/nonprod/terraform.tfvars#L19) and [prod](https://github.com/RaspberryPiFoundation/terraform/blob/5e4884626fc4d9fb84fa8b1083d7aab8d6f690c3/projects/editor-api/prod/terraform.tfvars#L18).

The CORS policies are set [here](https://github.com/RaspberryPiFoundation/editor-api/blob/08d36c56cf50c9da9e7ff6489179fa47bfe8f471/config/initializers/cors.rb) and just take the env var and separate the list on commas.

I'd like a more explicit set of CORS policies in the app similar to [badge-service](https://github.com/RaspberryPiFoundation/badge-service/blob/main/config/initializers/cors.rb). By doing these we'll also be able to wildcard specific subdomains to allow things like review apps to load the staging editor correctly!

### Other context

[CORP middleware](https://github.com/RaspberryPiFoundation/editor-api/blob/08d36c56cf50c9da9e7ff6489179fa47bfe8f471/lib/corp_middleware.rb) also utilises `ALLOWED_ORIGINS` to set headers required for accessing images from the web component. This could be altered too because it only uses the URLs without `https://` (a little ropey) but these are the API URLs (i.e. `staging-editor-api.raspberrypi.org` for staging and `editor-api.raspberrypi.org` for prod).

### Notes

Here's a list of URLs we need to include in the CORS policy

staging:

```
staging-editor-api.raspberrypi.org, (CORP middleware)
https://*.raspberrypi.org,

https://editor-standalone-eyq.pages.dev, (editor.raspberrypi.org CF pages build)
https://*.editor-standalone-eyq.pages.dev, (NEW - wildcard ones currently don't work so review apps don't load data)
https://block-to-text-alpha.pages.dev, (soon to be removed and deprecated so maybe a good candidate to leave in `ALLOWED_ORIGINS`)
https://projects-ui.pages.dev, (projects.raspberrypi.org CF pages build)
https://*.projects-ui.pages.dev, (NEW - wildcard ones currently don't work so review apps don't load data)

// REQUIRED FOR OAK
https://studio.edx.org,
https://preview.courses.edx.org,

[...]

PR #409, opened 2024-08-13
Allow for dynamic hosts in CORS origin config

grega:
## Status

- Closes https://github.com/RaspberryPiFoundation/editor-api/issues/405

## What's changed?

This extends the current origin-setting functionality (literal origins set via a comma-separated environment variable) and allows the env var string to also contain regular expressions. These allow the origins to match dynamic patterns (eg. `https://*.projects-ui.pages.dev`) and ensures that configuration still remains in the environment as we don't want to assume that anyone running this application will want the same allowed origins configured as us (other than for `localhost` for local and test environments only).

eg. instead of currently having to set the following in `ALLOWED_ORIGINS`:

```
https://foo.raspberrypi.org, https://bar.raspberrypi.org, https://foo.editor-standalone-eyq.pages.dev, https://bar.editor-standalone-eyq.pages.dev
```

The following can be set:

```
/https:\/\/(?:[a-z0-9-]+\.)?raspberrypi\.org$/, /https:\/\/.+\.editor-standalone-eyq\.pages\.dev$/
```

Associated updates in Terraform allow for the origins to match the spec in the [issue](https://github.com/RaspberryPiFoundation/editor-api/issues/405) (eg. the addition of wildcards / regex matching):

* https://github.com/RaspberryPiFoundation/terraform/pull/898

### DIFF
--- a/config/initializers/cors.rb
+++ b/config/initializers/cors.rb
@@ -7,14 +7,20 @@
 
 # Read more: https://github.com/cyu/rack-cors
 
-# Note the use of the $ at the end of the regexes below, this ensures
-# that the origins are matched exactly and not just a substring eg.
-# match: https://www.example.com
-# don't match: https://www.example.com.anotherdomain.com
+# fetch origins from the environment, these can be literal strings or regexes
+origins_array = ENV['ALLOWED_ORIGINS']&.split(',')&.map do |origin|
+  stripped_origin = origin.strip
+  if stripped_origin.start_with?('/') && stripped_origin.end_with?('/')
+    # convert to Regexp if the origin is wrapped in forward slashes eg. /https?:\/\/localhost(:[0-9]*)?$/
+    Regexp.new(stripped_origin[1..-2])
+  else
+    stripped_origin
+  end
+end || []
 
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
   allow do
-    # localhost and test domain
+    # localhost and test domains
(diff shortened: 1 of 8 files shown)


config/initializers/cors.rb
```diff
@@ -1,24 +1,31 @@
...
 
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
   allow do
-    origins origins_array
-    resource '*', headers: :any, methods: %i[get post patch put delete], expose: ['Link']
+    # localhost and test domain origins
+    if Rails.env.development?
+      origins(%r{https?://localhost([:0-9]*)$})
+    elsif Rails.env.test?
+      origins(%r{https?://localhost([:0-9]*)$}, %r{https?://www\.example\.com$})
```
conorriches, line 14: apologies if I'm not getting something here, do we want example.com in here?

conorriches (approved):
I've left a comment for something I'm just double checking on, but this looks great

config/initializers/cors.rb
grega: Good question! It was something I initially copied over from the reference badge-service implementation (I assume it's used in a few specs there) but don't think it's strictly needed here at all...

I'll drop it as it's not really adding anything to keep it around.

sra405 (approved):
👍 lgtm, just worth checking the corp middleware in particular in staging before deploying to prod

PR #425, opened 2024-09-04
232: Add mark as finished endpoint

danhalson:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/232

## What's changed?

- Adds a boolean `finished` to projects
- Adds a new endpoint  `/api/projects/front-chow-empty/toggle_finished` that toggles `finished`
- Restricted to students, on a remix of one of their projects only
- Returns the flag on a standard show, if the project is a remix associated with a school

- Run `db:migrate`

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -78,7 +78,7 @@ class Ability
       school_teacher_can_manage_project?(user:, school:, project:)
     end
     can(%i[read], Project, school_id: school.id, lesson: { visibility: %w[teachers students] })
-    can(%i[read toggle_finished], Project,
+    can(%i[read], Project,
         remixed_from_id: Project.where(user_id: user.id, school_id: school.id, remixed_from_id: nil).pluck(:id))
   end
   # rubocop:enable Metrics/AbcSize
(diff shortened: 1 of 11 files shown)


app/models/ability.rb
```diff
@@ -78,7 +78,7 @@ def define_school_teacher_abilities(user:, school:)
       school_teacher_can_manage_project?(user:, school:, project:)
     end
     can(%i[read], Project, school_id: school.id, lesson: { visibility: %w[teachers students] })
-    can(%i[read], Project,
+    can(%i[read toggle_finished], Project,
```
sra405, line 81: 🔧 its actually students that need to toggle this on any of their projects (i.e. with a `school_id` and `remixed_from_id`)

sra405: this will need cascading through the tests too 👍

sra405:
💭 it's probably worth extracting the `finished` parameter at the lesson level too (to surface "Ready to mark"). I believe it should be as simple as adding it to the lesson index [here](https://github.com/RaspberryPiFoundation/editor-api/blob/6e9293ae6a49b39b9ad306550593540a8eabfefb/app/views/api/lessons/index.json.jbuilder#L20-L27)

sra405 (approved):
👍 YYEESSSSSS!!!

PR #427, opened 2024-09-05
Reduce logged in students abilities

sra405:
## Status

- Related to https://github.com/RaspberryPiFoundation/editor-standalone/issues/335

## What's changed?

- Ensure from an API perspective students are unable to perform project/component calls that aren't assigned to a school (i.e. only assigned to their `user_id`)
- Remove capability to create a school or lesson
- Slight abilities refactor for readability

Limitations are in place on the frontend but this change ensures its also not possible from the backend too


app/models/ability.rb
```diff
@@ -4,22 +4,34 @@ class Ability
...
-    can :create, Component, project: { user_id: user.id, school_id: nil }
-
-    # Any authenticated user can manage their own projects.
-    can %i[read update destroy], Project, user_id: user.id
-    can %i[read update destroy], Component, project: { user_id: user.id }
+  def define_authenticated_non_student_abilities(user)
+    return if user&.student?
 
     # Any authenticated user can create a school. They agree to become the school-owner.
     can :create, School
```
danhalson, line 37: Shouldn't we ensure any abilities beyond this point are only given to a teacher or owner? else any one with an account could theoretically do these things - unlikely as that may be

sra405: Ah ok so we do want anyone to be able to do these things. We explicitly don't want students to be create a school (i.e. the school onboarding flow) or create projects/components that aren't attached to a school. Everyone else, teacher, owners, non-school members can do all those things

app/models/ability.rb
```diff
@@ -36,14 +48,14 @@ def initialize(user)
     # Any authenticated user can manage their own lessons.
     can %i[read create_copy update destroy], Lesson, user_id: user.id
```
sra405, line 49: 🔧 I do need to move this though, this is specifically a teacher/owner thing

danhalson, line 49: is this still meant to be here, we've excluded students but I wouldn't have thought we want a user that isn't a teacher or owner having the ability to do these things to a lesson though?

sra405: I did mull this one over for sure but it's a legacy idea thing around having a bank of lessons detached from a school similar to "My Projects". I mulled over deleting all the old tests and bits too. Maybe I'll explicitly add `school_id: nil`

sra405: there we go 😄 better than deleting well tested functionality we might use shortly

danhalson (approved):
LGTM

PR #431, opened 2024-09-10
add audit logging to class member table

### LINKED ISSUE #417: Remove class member endpoint
- [x] add endpoint to remove class members from the table
- [x] audit the class members table due to this being the first instance of an edit

conorriches:
Closes #417

adds audit logging to class member table, using school ID meta field

### DIFF
--- a/app/models/class_member.rb
+++ b/app/models/class_member.rb
@@ -13,7 +13,6 @@ class ClassMember < ApplicationRecord
   validate :student_has_the_school_student_role_for_the_school
 
   has_paper_trail(
-    if: ->(cm) { cm.school_class&.school_id },
     meta: {
       meta_school_id: ->(cm) { cm.school_class&.school_id }
     }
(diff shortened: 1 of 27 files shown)


app/models/class_member.rb
```diff
@@ -12,6 +12,13 @@ class ClassMember < ApplicationRecord
 
   validate :student_has_the_school_student_role_for_the_school
 
+  has_paper_trail(
+    if: ->(cm) { cm.school_class&.school_id },
```
danhalson, line 16: just wondering if you need the `if` here, since it will always have an associated `school_class` (and school by extension of that), and if it didn't we'd want that change and any others audited I think?

conorriches: I didn't know if there would guaranteed always be a `school_class` just from unfamiliarity, so I'll remove this `if` :)

conorriches: Done in [a276520](https://github.com/RaspberryPiFoundation/editor-api/pull/431/commits/a276520fb9d2299218eb2163d94bcad6e862fc68) :)

danhalson (approved):
LGTM

PR #433, opened 2024-09-12
Issues/355-add-school-members-endpoint

danhalson:
## Status

- Closes https://github.com/RaspberryPiFoundation/editor-standalone/issues/355

## What's changed?

- Adds a ClassMember::List endpoint
- Fleshes out SchoolTeacher::List endpoint (so it determines teachers from the school)

N/A

### DIFF
--- a/app/views/api/school_members/index.json.jbuilder
+++ b/app/views/api/school_members/index.json.jbuilder
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+json.array!(@school_members) do |school_member|
+  if school_member.respond_to?(:email) && school_member.email.present?
+    json.set! :teacher do
+      json.partial! '/api/school_teachers/school_teacher', teacher: school_member
+    end
+  else
+    json.set! :student do
+      json.partial! '/api/school_students/school_student', student: school_member
+    end
+  end
+end
--- a/spec/features/school_member/listing_school_members_spec.rb
+++ b/spec/features/school_member/listing_school_members_spec.rb
@@ -1,0 +1,137 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Listing school members', type: :request do
+  let(:headers) { { Authorization: UserProfileMock::TOKEN } }
+  let(:school) { create(:school) }
+  let(:owner) { create(:owner, school:) }
+  let(:teacher) { create(:teacher, school:) }
+  let(:students) { create_list(:student, 3, school:) }
+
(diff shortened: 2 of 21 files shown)


app/views/api/school_members/index.json.jbuilder
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+json.array!(@school_members) do |school_member|
+  if school_member.respond_to?(:email) && school_member.email.present?
```
sra405, line 4: ❓ trying to follow it through but is `school_member` the direct record from profile? Just wondering if there's another way of determining the role without having to do a join

sra405: ignore me, will see how easy it is to add a `type` into the profile response so we're not hacking around in different places/apps

spec/features/school_member/listing_school_members_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    stub_user_info_api_for(teacher)
+  end
+
+  it 'responds 200 OK' do
+    get("/api/schools/#{school.id}/members", headers:)
+    expect(response).to have_http_status(:ok)
+  end
+
+  it 'responds with the class members JSON array' do
```
sra405, line 28: think there's a couple of references to class members, will try and correct as I'm going

```suggestion
  it 'responds with the school members JSON array' do
```

spec/features/school_member/listing_school_members_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+  it 'does not include class members that belong to a different class' do
+    student = create(:student, school:)
+    different_class = create(:school_class, school:, teacher_id: teacher.id)
+    create(:class_member, school_class: different_class, student_id: student.id)
+
+    get("/api/schools/#{school.id}/members", headers:)
+    data = JSON.parse(response.body, symbolize_names: true)
+
+    expect(data.size).to eq(4)
+  end
```
sra405, line 116: ❓ a bit confused by this because it's a different school right but this seems to make a new student in the same school and they're not included in the response? 🤔 teachers/owners listing school members should see everyone

danhalson: you're quite right, I was a bit trigger happy on the old copy and paste

sra405 (dismissed):
👍 nice work. Loving the partials

sra405 (approved):
👍 👍

PR #445, opened 2024-10-03
Don't raise an error if graphql finds a missing repo

danhalson:
## Status

- Ready for review

## What's changed?

Prevents an error being raised if the github api query returns a NOT_FOUND error for the repo, since this can happen as a result of a race condition when the repo is first created.


sra405 (approved):
👍 yes!!

PR #450, opened 2024-10-23
Allow rejected schools to be verified

sra405:
## Status

Another common request is to remove `rejected_at` timestamps on schools that have gone back and changed aspects to allow them to be verified.


loiswells97 (approved):
Looks great, let's get this out there!

PR #451, opened 2024-10-24
404: Decrypt student password

danhalson:
## Status

- Related to https://github.com/RaspberryPiFoundation/editor-standalone/issues/404

## What's changed?

Passwords from editor-standalone will be encrypted for security, this decrypts them in editor-api

- Needs encryption key in terraform


spec/lib/tasks/classroom_management_spec.rb
loiswells97, line 1: I'm sure we already had something like this didn't we? Maybe you copied it and didn't remove the old one, or else git is just messing with us? 😅  Also, since you renamed the rake file, I guess this could be renamed too?

.circleci/config.yml
```diff
@@ -31,19 +31,20 @@ jobs:
       ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY: primary-key
       ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY: deterministic-key
       ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT: derivation-salt
+      EDITOR_ENCRYPTION_KEY: a1b2c3d4e5f67890123456789abcdef0123456789abcdef0123456789abcdef0
```
loiswells97, line 34: Does this need to go in the env.example as well?

danhalson: thanks! I must have removed it in a stash or something as it used to be there 😕

spec/lib/tasks/classroom_management_spec.rb
danhalson: Ahhh, this was git trying to fix a conflict - I'd moved the spec into a `/task` subdirectory, and didn't spot it getting recreated when I back-merged main

loiswells97 (approved):
Looks great! 👍

PR #452, opened 2024-10-24
Seed test data

loiswells97:
## Status

## Points for consideration:

- Security
- Performance

## What's changed?

### DIFF
--- a/lib/tasks/classroom_management.rake
+++ b/lib/tasks/classroom_management.rake
@@ -5,12 +5,12 @@ require_relative './seeds_helper'
 Rails.logger = Logger.new($stdout) unless Rails.env.test?
 
 # To override uuids call with:
-# `SEEDING_CREATOR_ID=00000000-0000-0000-0000-000000000000 rails classroom_management:seed_an_unverified_school`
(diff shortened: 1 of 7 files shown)


lib/tasks/classroom_management.rake
```diff
@@ -11,7 +11,7 @@ Rails.logger = Logger.new($stdout) unless Rails.env.test?
 # For students to match up the school needs to match with the school defined in profile (hard coded in the helper)
 
 namespace :classroom_management do
```
danhalson, line 13: wonder if there's a more appropriate name for this too now, an acronym like CEfE feels weird in a namespace so guess it could be CodeEditorForEducation or something more concise?

loiswells97: Maybe just `for_education`? Or is it weird to have a preposition in there? 🤔

danhalson: good call, I think it fits with the frontend page naming nicely and rolls off the tongue a bit better!

danhalson (approved):
LGTM

PR #456, opened 2024-11-01
Support multiple teachers per class

### LINKED ISSUE #463: Fix failing tests for multiple teacher support
There are a number of failing tests that need fixing before the API refactor enabling this can be merged.

Current in progress PR: https://github.com/RaspberryPiFoundation/editor-api/pull/456

Needs to be done in tandem with https://github.com/RaspberryPiFoundation/editor-standalone/issues/437

loiswells97:
## What's Changed?

- Added a `ClassTeacher` model which allows classes to have multiple teachers
- `Class` model now accepts nested attributes for `ClassTeacher`s
- `ClassMember` model has been renamed to `ClassStudent` to provide consistency with the approach for `SchoolStudent`, `SchoolMember` etc.
- `ClassMember` operations will now deal with both teachers and students in keeping with the `SchoolMember` module

closes #463


lib/concepts/class_member/operations/list.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ClassMember
+module ClassMember
```
KristinaDudnyk, line 3: I'm being curios, why do switch class here to a module?

loiswells97: Because we renamed the class_member table in the database to class_student, and also renamed the ClassMember model/class to ClassStudent. So we don't really care about creating instances of ClassMember anymore, just in using the methods it provides. These methods in some cases were taking account of both teachers and students, even though instances of the old ClassMember model could only represent students, which was confusing, hence the change.

KristinaDudnyk: Thanks for open anwer👏

KristinaDudnyk (approved):
Approved👍

app/controllers/api/projects_controller.rb
```diff
@@ -105,6 +106,10 @@ def school_owner?
       school && current_user.school_owner?(school)
     end
 
+    def class_teacher?(project)
+      project.lesson_id.present? && project.lesson.school_class.present? && project.lesson.school_class.teacher_ids.include?(current_user.id)
```
danhalson, line 110: do you need `project.lesson_id.present?`? I think that's implicit by `project.lesson` existing, since the relationship wouldn't exist otherwise

app/models/school_class.rb
```diff
@@ -2,12 +2,15 @@
 
 class SchoolClass < ApplicationRecord
   belongs_to :school
-  has_many :members, class_name: :ClassMember, inverse_of: :school_class, dependent: :destroy
+  has_many :students, class_name: :ClassStudent, inverse_of: :school_class, dependent: :destroy
+  has_many :class_teachers, class_name: :ClassTeacher, inverse_of: :school_class, dependent: :destroy
```
danhalson, line 6: why use `:students` and then `:class_teachers` instead of `:teachers` or vice versa? going with one or the other would be nice for consistency

danhalson (changes requested):
Looks great, just a few fairly minor comments

app/models/school_class.rb
loiswells97: Ah yes, that would simplify things, thanks 👍

app/models/project.rb
```diff
@@ -96,11 +96,22 @@ def user_has_a_role_within_the_school
     errors.add(:user, msg)
   end
 
-  def user_is_a_member_or_the_owner_of_the_lesson
+  def user_is_class_teacher_or_student
     # TODO: Revisit the case where the lesson is not associated to a class i.e. when we build a lesson library
```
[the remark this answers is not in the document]
loiswells97: I haven't really thought much about the case where the lesson is not associated to a class, so might be worth leaving on for now, if only as a reminder that it might need more thought

app/controllers/api/projects_controller.rb
loiswells97: This was more to be on the safe side to cover off the cases where it's not a school project

app/views/api/school_classes/index.json.jbuilder
```diff
@@ -1,16 +1,23 @@
...
     :school_id,
-    :teacher_id,
     :name,
     :created_at,
     :updated_at
   )
 
-  json.teacher_name(teacher&.name)
+  json.teachers(teachers) do |teacher|
+    if teacher.present?
```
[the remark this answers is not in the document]
loiswells97: 👍

danhalson (approved):
Looks great! and well done, this was a prickly bit of work :)

PR #459, opened 2024-11-11
Don't try and encrypt password if none passed

danhalson:
## Status

- Ready for review

## What's changed?

- Only try and decrypt the password if a password is passed, since resetting and updating student details use the same endpoint 

N/A


loiswells97 (approved):
Looks great, thanks for looking into this 👍

PR #470, opened 2024-12-02
Renaming lesson project and student remixes

loiswells97:
## Status

## Points for consideration:

- Security
- Performance

## What's changed?

### DIFF
--- a/lib/concepts/lesson/operations/update.rb
+++ b/lib/concepts/lesson/operations/update.rb
@@ -8,12 +8,27 @@ class Lesson
         response[:lesson] = lesson
         response[:lesson].assign_attributes(lesson_params)
         response[:lesson].save!
+        if lesson_params[:name].present?
+          rename_lesson_project(lesson: response[:lesson], name: lesson_params[:name])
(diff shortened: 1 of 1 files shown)


lib/concepts/lesson/operations/update.rb
```diff
@@ -8,13 +8,32 @@ def call(lesson:, lesson_params:)
...
+      def rename_lesson_project(lesson:, name:)
+        return unless lesson.project
+
+        lesson.project.update!(name:)
+      end
+
+      def rename_lesson_remixes(lesson:, name:)
+        return unless lesson.project
+
+        lesson_remixes = Project.where(remixed_from_id: lesson.project.id)
```
danhalson, line 32: would it be worth considering a transaction here? just to ensure we don't end up with some renamed and some not which might be messy to unpick

loiswells97: ooh yes I hadn't thought of that 👀

loiswells97: Okay, that should be done 👍

danhalson (approved):
Looks great!

PR #473, opened 2024-12-10
Include project when lesson is copied

loiswells97:
## Status

## Points for consideration:

- Security
- Performance

## What's changed?

### DIFF
--- a/lib/concepts/lesson/operations/create_copy.rb
+++ b/lib/concepts/lesson/operations/create_copy.rb
@@ -19,9 +19,24 @@ class Lesson
 
       # TODO: copy projects
       def build_copy(lesson, lesson_params)
-        copy = Lesson.new(parent: lesson, name: lesson.name, description: lesson.description)
-        copy.assign_attributes(lesson_params)
-        copy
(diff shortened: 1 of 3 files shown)


lib/concepts/lesson/operations/create_copy.rb
```diff
@@ -17,11 +17,29 @@ def call(lesson:, lesson_params:)
...
+        lesson_copy.assign_attributes(lesson_params)
+
+        project_params = { name: lesson_copy.name, user_id: lesson_params[:user_id], lesson_id: lesson_copy.id }
+        lesson_copy.project = build_project_copy(lesson.project, project_params)
+
+        lesson_copy
+      end
+
+      def build_project_copy(project, project_params)
+        project_attributes = project.attributes.except('id', 'identifier').merge(project_params)
```
danhalson, line 31: should we be avoiding `created_at` and `updated_at` here and for the lesson? it seems like we can end up with some weird values, until updates occur otherwise

loiswells97: Ah yes, thanks Dan, I've excluded these now

danhalson (approved):
Looks great!

PR #474, opened 2024-12-16
Add my classes filter to classes index endpoint

loiswells97:
## What's changed?

- Only return the classes for which the user is the teacher if `my_classes` query param is set


sra405 (approved):
👍

PR #482, opened 2025-01-24
Distinguish media types

loiswells97:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/384

## What's changed?

- Added `videos` and `audio` attributes to the `project` model
- Added logic to distinguish between media types in the project upload job
- Refactored `ProjectImporter` to account for videos and audio files as well as images
- Surfaced videos and audio files to the frontend alongside images


app/jobs/upload_job.rb
```diff
@@ -86,30 +86,55 @@ def load_projects_data(locale, repository, owner)
...
+  def categorize_files(files, project_dir, locale, repository, owner)
+    categories = {
+      components: [],
+      images: [],
+      videos: [],
+      audio: []
+    }
+
+    files.each do |file|
+      next if file.name == 'project_config.yml'
```
danhalson, line 119: is this necessary since you're calling reject for the config above?

app/jobs/upload_job.rb
```diff
@@ -86,30 +86,55 @@ def load_projects_data(locale, repository, owner)
...
+    categorized_files = categorize_files(files, project_dir, locale, repository, owner)
+
+    { **proj_config, locale:, **categorized_files }
+  end
+
+  def validate_directory_structure(data)
+    raise InvalidDirectoryStructureError, 'The directory structure is incorrect and the job can\'t be processed.' unless data.respond_to?(:entries)
+  end
+
+  def categorize_files(files, project_dir, locale, repository, owner)
```
danhalson, line 110: I think this is an excellent refactor 😄

app/models/filesystem_project.rb
```diff
@@ -4,33 +4,54 @@
 
 class FilesystemProject
   CODE_FORMATS = ['.py', '.csv', '.txt', '.html', '.css'].freeze
-  IMAGE_FORMATS = ['.png', '.jpg', '.jpeg', '.webp'].freeze
   PROJECTS_ROOT = Rails.root.join('lib/tasks/project_components')
 
   def self.import_all!
     PROJECTS_ROOT.each_child do |dir|
       proj_config = YAML.safe_load_file(dir.join('project_config.yml').to_s)
```
danhalson, line 11: could we put `project_config.yml` into a variable?

app/jobs/upload_job.rb
```diff
@@ -86,30 +86,55 @@ def load_projects_data(locale, repository, owner)
...
-    proj_config = {}
-
     data = project_dir.object
-    raise InvalidDirectoryStructureError, "The directory structure is incorrect and the job can't be processed." unless data.respond_to?(:entries)
+    validate_directory_structure(data)
 
-    data.entries.each do |file|
-      if file.name == 'project_config.yml'
-        proj_config = YAML.safe_load(file.object.text, symbolize_names: true)
+    proj_config_file = data.entries.find { |file| file.name == 'project_config.yml' }
```
danhalson, line 92: can we also stick `project_config.yml` in a variable here

app/jobs/upload_job.rb
loiswells97: This was only because `rubocop` was kicking off about cyclomatic complexity 🤣

app/jobs/upload_job.rb
loiswells97: Ah I'd missed this, thanks! 🗑️

danhalson (approved):
Looks great!

PR #490, opened 2025-02-06
Include `finished` flag in remix index

### LINKED ISSUE #485: Surface `finished` flag when listing student remixes

loiswells97:
## Status

closes #485

### DIFF
--- a/app/views/api/projects/remixes/index.json.jbuilder
+++ b/app/views/api/projects/remixes/index.json.jbuilder
@@ -7,6 +7,7 @@ json.array!(@projects_with_users) do |project, user|
     :project_type,
     :name,
     :user_id,
+    :finished,
     :updated_at,
     :last_edited_at
   )


danhalson (dismissed):
lgtm

db/migrate/20250207095204_remove_finished_column_from_projects.rb
```diff
@@ -0,0 +1,19 @@
+class RemoveFinishedColumnFromProjects < ActiveRecord::Migration[7.1]
+  def up
+    Project.all.each do |project|
+      if project.school_project.present?
+        project.school_project.update!(finished: project.finished)
```
danhalson, line 5: `project.finished` hasn't been used anywhere yet, so you shouldn't need this

danhalson: ah ignore that, are you intending to add a new `school_project` for each project, as I think we'll need that? I think the logic needs adjusting so that it's checking if `school_project.blank?` and `school_id.present?`, and creating a new record if so?

app/models/project.rb
```diff
@@ -2,6 +2,7 @@
 
 class Project < ApplicationRecord
```
danhalson, line 3: we'll also need to ensure a SchoolProject is created every time a Project is, we could either do this in a hook like `after_create`, though the logic to ensure it only happens for schools might be trickier to implement here, or do it in Projects concept, and maybe add a validation (though that might negate avoiding using a hook since the logic will be similarish to validate)

db/migrate/20250207095204_remove_finished_column_from_projects.rb
loiswells97: I think I just added this in for the sake of completeness, in case we ever roll these migrations back for some reason

app/controllers/api/school_projects_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Api
+  class SchoolProjectsController < ApiController
+    before_action :authorize_user
+    load_and_authorize_resource :project
+
+    def show_finished
```
danhalson, line 8: this is semantics, but I wonder if `get_finished` would be a better name?

app/models/project.rb
```diff
@@ -10,17 +10,21 @@ class Project < ApplicationRecord
   has_many_attached :images
   has_many_attached :videos
   has_many_attached :audio
+  has_one :school_project, dependent: :destroy
 
   accepts_nested_attributes_for :components
 
   before_validation :check_unique_not_null, on: :create
+  before_validation :create_school_project_if_needed
```
danhalson, line 18: I think it would be safer to use `after_save` for this, so we can ensure the record has passed validation and has actually been saved into the database

loiswells97: Hmmm, I guess in that case we wouldn't be able to validate the school project in the same way as we are currently 🤔

danhalson (approved):
Looks great! sorry for unhelpful comments

PR #492, opened 2025-02-10
Add libyaml-dev to Dockerfile

chrisroos:
This is required to build the psych Gem. Prior to this commit, running `docker compose build` would fail with:

```
65.20 Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
65.20
65.20     current directory: /usr/local/bundle/gems/psych-5.1.2/ext/psych
65.20 /usr/local/bin/ruby extconf.rb
65.20 checking for yaml.h... no
65.20 yaml.h not found
65.20 *** extconf.rb failed ***
```

I don't understand why I'm seeing this error now because there don't appear to be any changes in the Dockerfile or docker-compose.yml files that could've caused it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ FROM base AS builder
 WORKDIR /app
 RUN apt-get update \
   && apt-get install --yes --no-install-recommends \
-  build-essential libpq-dev libxml2-dev libxslt1-dev git \
+  build-essential libpq-dev libxml2-dev libxslt1-dev git libyaml-dev \
   firefox-esr python2-dev \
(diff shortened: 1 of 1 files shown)


danhalson (approved):
Hi Chris, we're only building with Apple Silicon these days and haven't tested anything else, are you building on an Intel based Mac perhaps, or a different architecture?

Either way change looks good to me, so approving :)

chrisroos:
Thanks @danhalson :-)

I'm on an Intel Linux machine but I don't think that's the problem as @floehopper sees the same error when running on an Apple Silicon Mac.

PR #502, opened 2025-02-17
Fix for attaching media in editor builds

danhalson:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/438

## What's changed?

In some scenarios the blob is not being created properly on attach, which is causing a foreign key constraint, this change explicitly creates the blob prior to attaching it.

- Try re-running one of the failed jobs


loiswells97 (approved):
Looks great, thanks for looking into this! 👍

PR #516, opened 2025-03-03
Allow class members create route to take teachers as well as students

### LINKED ISSUE #464: Allow class member create route to receive both students and teachers
Currently the class member create route is only set up to handle adding students to the class. Change this so that teachers and students can be added to the class alongside one another in the same request.

This is blocking https://github.com/RaspberryPiFoundation/editor-standalone/issues/438

loiswells97:
## What's Changed?

- Updated class member `create` and `create_batch` routes to accept both students and teachers
- `ClassMembers::Create` operation now deals with both students and teachers

closes #464


app/controllers/api/class_members_controller.rb
```diff
@@ -20,9 +20,16 @@ def index
     end
 
     def create
-      student_ids = [class_member_params[:student_id]]
-      students = SchoolStudent::List.call(school: @school, token: current_user.token, student_ids:)
-      result = ClassMember::Create.call(school_class: @school_class, students: students[:school_students])
+      user_ids = [class_member_params[:user_id]]
+      user_type = class_member_params[:type]
+      if %w[teacher owner].include?(user_type)
```
danhalson, line 25: just wondering...if we eventually have owners that aren't teachers, would we want them as class members I wonder? the permissions currently don't lend themselves to that iirc (happy to be corrected though)

loiswells97: Hmmm yes, that's a good point. I think as long as we're definitely sending `teacher` rather than `owner` from the frontend in the case where the user is both an owner and teacher then I think we can safely get rid of the owner option here

app/controllers/api/class_members_controller.rb
```diff
@@ -33,8 +40,15 @@ def create
     end
 
     def create_batch
-      students = SchoolStudent::List.call(school: @school, token: current_user.token, student_ids: create_batch_params)
-      result = ClassMember::Create.call(school_class: @school_class, students: students[:school_students])
+      teacher_objects = create_batch_params.select { |user| user[:type] == 'teacher' }
+      student_objects = create_batch_params.select { |user| teacher_objects.pluck(:user_id).exclude?(user[:user_id]) }
```
danhalson, line 44: maybe I'm missing something and this has changed now you're not including owners, but can't you just select students by type?

danhalson (approved):
LGTM

PR #519, opened 2025-03-03
Add example Scratch project for Experience CS 

floehopper:
## Points for consideration:

Do not merge until #513 has been merged and deployed. This will ensure that existing clients of the API are unable to "see" this example Scratch project.

## What's changed?

This adds a simple example project of type "scratch". For the moment we're only planning to store metadata for Scratch projects in the `editor-api` database, so no components are included.

The seed data in production-like environments will need to be updated so this new project is added.

### DIFF
--- a/lib/tasks/project_components/experience_cs_example/project_config.yml
+++ b/lib/tasks/project_components/experience_cs_example/project_config.yml
@@ -1,0 +1,4 @@
+NAME: "Example"
+IDENTIFIER: "experience-cs-example"
+TYPE: "scratch"
+COMPONENTS: []


loiswells97 (commented):
Looks like we're not using the `COMPONENTS` key anymore as we could just work out what the files were meant to be from the content of the directory

chrislo (approved):

chrisroos:
> Looks like we're not using the `COMPONENTS` key anymore as we could just work out what the files were meant to be from the content of the directory

Thanks, @loiswells97. I think you're correct. It looks as though the importer code has been updated at some point but not the corresponding project_config.yml files. The [README](https://github.com/RaspberryPiFoundation/editor-api/blob/b54f450d0eda5e62892029704441291e75d7fdee/lib/tasks/README.md) also seems to be out of sync with the code as it appears to predate the addition of html project types. I'm not going to address this tidy-up as part of this PR.

PR #527, opened 2025-03-12
Remove projects:create_all from release

danhalson:
## Status

Ready for review

## What's changed?

Removes projects:create_all from release process, which could be resetting default projects

N/A


loiswells97 (approved):
✅

PR #528, opened 2025-03-12
fix owners being added to class

loiswells97:
## What's changed?

- Fixed ability to add owners to a class
- Fixed owner appearing with teacher icon when first added to a class


adrian-rpf (approved):
Looks good.

PR #531, opened 2025-03-20
Add project context route

loiswells97:
## Status

closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/479

## What's changed?

- Added project context route to allow apps to request contextual data about a project. This will return the school, class and lesson ids for the project.
- Stopped students from viewing their work if the associated lesson is no longer visible to students


app/controllers/api/projects/remixes_controller.rb
```diff
@@ -14,6 +14,7 @@ def index
 
       def show
         @project = Project.find_by!(remixed_from_id: project.id, user_id: current_user&.id)
+        authorize! :show, @project
```
danhalson, line 17: just a thought, we're using cancancan a bit inconsistently in places, and I wonder if `load_and_authorize_resource :project, only: [:show]` could be a cleaner approach?

app/controllers/api/projects_controller.rb
```diff
@@ -52,6 +52,10 @@ def destroy
       head :ok
     end
 
+    def show_context
+      render :context, formats: [:json]
```
danhalson, line 56: maybe a comment explaining the purpose would be a good? as I don't think it's easy to understand why we're doing this just by reading the code...

danhalson (changes requested):
I think this is great, just a couple of minor suggestions

app/controllers/api/projects/remixes_controller.rb
loiswells97: would that still work with the way we're loading up the remix project?

danhalson (approved):
LGTM

PR #534, opened 2025-04-10
WIP: Optionally include Scratch projects in GET /remix

chrisroos:

PR #536, opened 2025-04-16
Create and show remixes for scratch projects

floehopper:
This extends the `project_type` param mechanism introduced in https://github.com/RaspberryPiFoundation/editor-api/pull/513 [into the `Api::Projects::RemixesController`](https://github.com/RaspberryPiFoundation/editor-api/commit/58875819cfebbfbd0f3a999271d78ae7544e1f3d) so that [the `experience-cs` app][1] can create a remix of a Scratch project and then show the remix. See [this pull request][2] for more details.

It also includes some preparatory refactoring [to extract a `Project.only_scratch` scope](https://github.com/RaspberryPiFoundation/editor-api/commit/00304473f251caca176e0de0977b9467ed4cf381) to reduce duplication. And then some subsequent refactoring [to extract a `ProjectScoping` controller concern and `Current.project_scope`](https://github.com/RaspberryPiFoundation/editor-api/commit/7711490ebe4962e88617081af3286769fee799d5) to further reduce duplication and simplify the code. The latter should make it easier to make similar changes to other controller actions in the future as we add functionality to `experience-cs`. It should also make it easier to change the mechanism we use for this, e.g. using an HTTP header instead of the `project_type` param.

Supersedes #534.

[1]: https://github.com/RaspberryPiFoundation/experience-cs
[2]: https://github.com/RaspberryPiFoundation/experience-cs/pull/290

### DIFF
--- a/app/controllers/api/projects/remixes_controller.rb
+++ b/app/controllers/api/projects/remixes_controller.rb
@@ -36,12 +36,16 @@ module Api
       private
 
       def project
-        @project ||= Project.find_by!(identifier: params[:project_id])
+        @project ||= project_scope.find_by!(identifier: params[:project_id])
       end
 
       def load_and_authorize_remix
-        @project = Project.find_by!(remixed_from_id: project.id, user_id: current_user&.id)
+        @project = project_scope.find_by!(remixed_from_id: project.id, user_id: current_user&.id)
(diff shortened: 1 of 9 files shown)


chrisroos (approved):
Looks good to me, @floehopper!

PR #541, opened 2025-05-06
Add example to `SchoolTeacher::Invite` spec to diagnose flakey example

floehopper:
The example just before this one keeps failing in the CircleCI build, but I haven't been able to replicate it locally. It's hard to work out what's going on, because `SchoolTeacher::Invite#call` swallows any exception. I'm hoping that this new assertion _might_ fail for the same reason, but then display the exception message to give us more clues.


floehopper:
This was addressed in https://github.com/RaspberryPiFoundation/editor-api/commit/61664b72b8eff577f294f0bcedb2b3e7c59b1a0b as part of #561. Closing.

PR #551, opened 2025-05-25
Allow ExperienceCS admins to create public projects v2

floehopper:

PR #552, opened 2025-05-25
Allow Experience CS admins to manage public projects

floehopper:
Supersedes #549 & #551


chrisroos (dismissed):
I've been through this in detail and subsequently chatted to @floehopper about it. We agreed that it's probably better to restrict the `experience-cs-admin` permissions further and only allow them to CRUD Projects where the user_id is nil to hopefully avoid admins updating user-owned projects. James is going to work on that change but I'm going to approve this in the meantime so that it's ready to merge once that change has been made.

sebjacobs (approved):

PR #553, opened 2025-05-25
Fix authorization in Api::ProjectsController

floehopper:
Previously this spec was not representative of how the API was being used in practice and it was only really working by accident. The `PUT` requests in the examples were using the `Project#id` rather than the `Project#identifier` as they should have been.

The spec was working even though the `Api::ProjectsController#load_project` before action was *not* finding
the project, because the `load_and_authorize_resource` before action was then finding the project using `Project.find(params[:id])` and setting the `@project` instance variable used by the rest of the logic in the controller.

However, `editor-standalone` [uses the project identifier as the resource ID in the URL path][1], so this spec was completely unrepresentative of how the endpoint was really being used.

In this commit I've changed the examples to use the `Project#identifier` as the resource ID in the `PUT` requests. This means that the `Api::ProjectsController#load_project` before action now finds the project by identifier and since it sets the `@project` instance variable the `load_and_authorize_resource` before action no longer attempts to load the resource, it just does the authorize step using the already loaded project.

In order to make this work reliably, I had to explicitly set the `Project#locale` to [one of the fallback locales in `ProjectLoader`][2], i.e. 'en' or `nil`. Otherwise, the random locale selected in the project factory [3] meant that sometimes the `load_project` before action (which uses the `ProjectLoader` did not find the project and the `load_and_authorize_resource` before action raised an `ActiveRecord::RecordNotFound` resulting in a 404 Not Found.

Now the spec is no longer making use of the randome locale from the factory, I've taken the opportunity to add some new examples demonstrating the behaviour when the project has different locales. This effectively demonstrates that `load_project` is wired up to `ProjectLoader` correctly.

As an aside, I'm not convinced that having the factory select a locale at random is a good idea. I found it very confusing when it led to undeterministic specs. However, I'll leave that for another time.

To make it clearer the `load_project` before action is where we want the project to be loaded, I've changed the `load_and_authorize_resource` before action to [`authorize_resource`][1], so the CanCanCan authorization uses the project found by the `load_project` before action.

However, this meant that if the project was *not* found by the `load_project` before action an exception was raised in `Project::Update.call` resulting in a `422 Unprocessable Entity` response with the following error message:

    Error persisting changes: undefined method `components' for nil:NilClass

To fix this I'm now raising an `ActiveRecord::RecordNotFound` exception in the `load_project` before action if no project is found. This results in the expected `404 Not Found` response.

I think there's a strong case to be made the this exception raising behaviour should be added to `ProjectLoader#load`. However, that's a bigger change with a lot more risk, so I'm going to leave that for now.

[1]: https://github.com/RaspberryPiFoundation/editor-standalone/blob/1d4375635cb6890794732072d608dbd4b05b3bb0/src/utils/apiCallHandler/projects.js#L16
[2]: https://github.com/RaspberryPiFoundation/editor-api/blob/b4bd337d09a88b1f41ecdc13136f7d11da0dcf89/lib/project_loader.rb#L8
[3]: https://github.com/RaspberryPiFoundation/editor-api/blob/b4bd337d09a88b1f41ecdc13136f7d11da0dcf89/spec/factories/project.rb#L9
[4]: https://github.com/CanCanCommunity/cancancan/blob/3.4.0/docs/controller_helpers.md#authorize_resource-load_resource-load_and_authorize_resource


floehopper:
@chrisroos

> I've been looking through the history and think I have a better idea of what might've happened:

Amazing detective work! That does give me more confidence.

I've now rebased this against `main` which forced me to add [Move let into private method in Updating a project spec](https://github.com/RaspberryPiFoundation/editor-api/pull/553/commits/5fda5a12da59c9436d405cef23cf17ca07b8be89) to avoid triggering a Rubocop cop violation in one of the later commits.

chrisroos (approved):
Looks good to me, @floehopper!

PR #556, opened 2025-05-29
605: Add user origin to school dashboard, fix relationships

danhalson:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/605

## Points for consideration:

- Security
- Performance

## What's changed?

- Adds the user_origin to the school dashboard
- Fixes the relationships between models since class members were split out

N/A

### DIFF
--- a/app/dashboards/school_class_dashboard.rb
+++ b/app/dashboards/school_class_dashboard.rb
@@ -46,10 +46,7 @@ class SchoolClassDashboard < Administrate::BaseDashboard
   # FORM_ATTRIBUTES
   # an array of attributes that will be displayed
   # on the model's form (`new` and `edit`) pages.
-  FORM_ATTRIBUTES = %i[
-    name
-    school
-  ].freeze
+  FORM_ATTRIBUTES = %i[].freeze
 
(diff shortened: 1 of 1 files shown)


app/fields/enum_field.rb
danhalson, line 1: Without this, administrate tries to treat enums as select fields and will error, since the data is not in the expected format

app/controllers/admin/projects_controller.rb
danhalson, line 1: A blanket scope was added to all projects that limits them to internal only, this is problematic since it prevents users from seeing projects related to lessons, so this applies the scope only to the main index action.

app/dashboards/class_student_dashboard.rb
danhalson, line 1: The student and teacher dashboards only show the profile id owing to complications around pulling in user info from profile, this is out of scope of this issue and will need to be expanded in the future...this at least allows things to render, and a dev can look up the user_id if need be.

loiswells97 (approved):
Looks great ✅

PR #558, opened 2025-06-03
Remove default scope excluding scratch projects

floehopper:
This removes the default scope on `Project` that hid scratch projects by default which was originally added in #513 to support work on the Experience CS project while leaving Code Editor, Code Editor for Education, etc, unaffected.

The Code Editor team are going to add protection to `editor-standalone` and `editor-ui` so that they show an error message and/or redirect the user if they are unintentionally used to load a scratch project. **Once this protection is in place**, we'll be able to apply the changes in this pull request to remove the default scope so that scratch projects are no longer hidden. This is necessary to support ongoing integration between Experience CS and Code Editor for Education.

As we agreed, I haven't made any changes to the GraphQL endpoint. In any case, I _think_ any changes to filter on `project_type` would need to be made on the client side.

I decided not to fully revert all the changes, because I think extracting the project type string literals into `Project::Types::PYTHON`, `Project::Types::HTML` & `Project::Types::SCRATCH` is still a useful change. And because of that and some later changes, I wasn't able to simply revert a couple of the commits, so I constructed those commits manually.

I decided to remove the `scratch_only` scoping mechanism as well as the default scope, because I don't think we really need it in `experience-cs` - I think we can probably replace it with something similar to what the Code Editor team are going to do, i.e. display an error and/or redirect if the app is unintentionally used to load a non-scratch project. Having said that, there's no need to wait for us to make that change before merging this pull request.


.rubocop.yml
```diff
@@ -20,7 +20,7 @@ GraphQL/ObjectDescription:
     - app/graphql/types/query_type.rb
 
 RSpec/NestedGroups:
-  Max: 5
+  Max: 4
```
loiswells97, line 23: Just wondering the reason for the reduction here?

app/models/project.rb
```diff
@@ -32,12 +32,7 @@ module Types
   validate :project_with_school_id_has_school_project
   validate :school_project_school_matches_project_school
 
-  default_scope -> { where.not(project_type: Types::SCRATCH) }
-
   scope :internal_projects, -> { where(user_id: nil) }
-  scope :only_scratch, lambda { |only_scratch|
-    only_scratch ? unscoped.where(project_type: Project::Types::SCRATCH) : self
-  }
```
loiswells97, line 40: I thought the `only_scratch` stuff was still needed on the ExCS side? No worries if I misunderstood or that's changed

lib/tasks/projects.rake
```diff
@@ -54,7 +54,7 @@ namespace :projects do
     ]
     projects.each do |attributes|
       identifier = attributes[:identifier]
-      project = Project.only_scratch(true).find_by(attributes.slice(:identifier, :locale))
+      project = Project.find_by(attributes.slice(:identifier, :locale, :project_type))
```
loiswells97, line 57: Is there a reason to include `project_type` in here? Projects should be have unique identifiers within the locale.

.rubocop.yml
floehopper: We increased it from 4->5 in [the commit which added the scratch-only option](https://github.com/RaspberryPiFoundation/editor-api/commit/56022c6f1dbea4ea1f94c95497aab3971e7c3d2d), so it seemed right to revert that change since the higher limit isn't needed anywhere else.

floehopper:
@loiswells97

> Just a few questions on here, but apart from that, looking great and keen to get this in

Many thanks for your reviewing. I've responded to your questions inline. Let me know if you need any further clarification or want me to change anything before merging.

Also can you confirm that you now have the measures we discussed in place in editor-standalone & editor-ui to handle the case where someone tries to load a scratch project into the Code Editor? Just wanted to triple-check!

lib/tasks/projects.rake
loiswells97: I'm more thinking that removing it would mean that it wouldn't overwrite projects of a different type with the same identifier - at the moment it would overwrite unless the existing project is a scratch project. I think it depends how and where this is being used as to how much it matters, but removing it would be my instinct.

app/models/project.rb
loiswells97: Great 👍

lib/tasks/projects.rake
floehopper: > I'm more thinking that removing it would mean that it wouldn't overwrite projects of a different type with the same identifier - at the moment it would overwrite unless the existing project is a scratch project.

Apologies - I'm probably misunderstanding something in what you're saying.

The `Project.find_by(attributes.slice(:identifier, :locale, :project_type))` call will only find an existing record (and potentially update it) if *all three* attributes (`identifier`, `locale` & `project_type`) match. So it won't match a project that has the same `identifier` & `locale` but has a different `project_type` and thus it won't update it.

I agree that in practice, because the combination of `identifier` & `locale` has to be unique, strictly speaking it is _unnecessary_ to also find by `project_type`. However, I can't see how also finding by `project_type` would ever do the wrong thing, i.e. overwrite a project that we don't want it to. Are you seeing a way that can happen? If so, can you take me through a concrete step-by-step example so I understand? I don't want to merge this if that's a possibility.

Another way to think about this change is that previously the following call only considered Scratch projects, because of the `Project.only_scratch(true)` call:

```ruby
project = Project.only_scratch(true).find_by(attributes.slice(:identifier, :locale))
```

In the new version, including the `project_type` attribute in the `find_by` and given all the hard-coded project hashes above all have a `project_type` of Scratch, means it will only ever find Scratch projects, i.e. the behaviour is effectively the same as before:

```ruby
project = Project.find_by(attributes.slice(:identifier, :locale, :project_type))
```

> I think it depends how and where this is being used as to how much it matters, but removing it would be my instinct.

This rake task was introduced specifically to create/update seed Scratch projects for the ExCS project in the Heroku deployment release phase. We have also used it in development to create/update those same seed/projects. The Learning Team now have the capability to create Scratch starter projects in the ExCS admin UI, so very soon we'll be removing the task completely.

loiswells97 (approved):
Looks good, let's get this out there 🚢

lib/tasks/projects.rake
floehopper: Ah, I see - thank you! 👍

floehopper:
Rebasing on `main` and force-pushing in preparation for merging.

PR #559, opened 2025-06-05
Make projects index action accessible to anonymous users

floehopper:


floehopper:
@sebjacobs I also considered adding a 'public' action to 'Api::ProjectsController' to pretty much copy the 'index' action but only include public projects and only authorize access to ExCS admins. This would avoid changing existing behaviour at the cost of complicating the system potentially unnecessarily.

PR #566, opened 2025-07-08
Use Identifiable#current_user in API controllers

floehopper:
Previously these were using the `@current_user` instance variable. However, this was working somewhat by accident, because `ApiController#authorize_user` (called by a `before_action`) was calling `Identifiable#current_user` and thus setting `@current_user`.

I think it's safer, more idiomatic and less confusing to have an explicit call to the memo-ized `Identifiable#current_user`. Especially as that's what we're already using in all the other controllers and even in other places in these controllers!


sebjacobs (approved):

floehopper:
@adrian-rpf This addresses one of the issues I mentioned when we chatted the other day. I'd welcome your feedback.

PR #567, opened 2025-07-08
Remove unused API endpoints and unused stubbing of Profile API

floehopper:
I started making these changes when I noticed that a bunch of [stubbing](https://github.com/RaspberryPiFoundation/editor-api/blob/7cf9faf02696d7eb14afb3fe1b3e119f99ca983f/spec/support/profile_api_mock.rb) of [`ProfileApiClient`](https://github.com/RaspberryPiFoundation/editor-api/blob/7cf9faf02696d7eb14afb3fe1b3e119f99ca983f/lib/profile_api_client.rb) methods was not needed.

However, I then noticed that a bunch of `ProfileApiClient` methods are no-ops, because they don't make any use of [`ProfileApiClient.connection`](https://github.com/RaspberryPiFoundation/editor-api/blob/7cf9faf02696d7eb14afb3fe1b3e119f99ca983f/lib/profile_api_client.rb#L186-L197) which is the only code that actually makes any HTTP requests to the Profile API. This in turn led me to remove the API endpoints whose sole purpose was to call these no-op methods along with all the associated code and specs.

I can't see any reason to keep unused code in the app - it just makes the codebase harder to navigate for no good reason. If we envisage implementing these endpoints in the future and we're worried about having to write the same code again, we can easily recover it from the git history.


chrisroos (approved):
This all looks good to me, @floehopper!

I've also double checked CEfE in staging and can't see any way in the UI to do any of the 3 operations you've removed:

- Add an owner
- Remove an owner
- Remove a teacher

Which provides further evidence that this code isn't being exercised.

PR #572, opened 2025-07-30
Enable reseed on test API

loiswells97:
Removes the env check so we can run this on the test API which has Rails env set to production. There are other safeguards in place such as an API key and host check that should prevent this from actually running on production.


danhalson (approved):
LGTM

PR #576, opened 2025-07-31
Reenable reseed rake tasks

loiswells97:
## Status

- Related to https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/764

## What's changed?

Rake tasks can only be run once per process, so we need to reenable prior to running so they can be run multiple times, which is necessary in this context since it's an api call.

N/A


danhalson (dismissed):
LGTM

PR #579, opened 2025-08-06
Fix flakey for education spec

chrislo:
We've been seeing a flakey test like

    Failures:

    1) for_education :seed_a_school_with_lessons_and_students creates lessons with projects
       Failure/Error: expect(lesson.length).to eq(2)

         expected: 2
              got: 4

         (compared using ==)
       # ./spec/lib/tasks/for_education_spec.rb:74:in `block (3 levels) in <top (required)>'

when running this spec file in CI.

I haven't been able to puzzle out what's going on but I suspect there's some kind of race-condition happening. The task under test relies on some hard-coded specific identifers for the user and school and these are accessed via a global variable. I wonder if it's possible that these are leaking between tests and the same school is being operated on twice before the expectations run.

In any case, as far as I can tell this task will add lessons to the school each time it is run (see `SeedsHelper#create_lessons`) so I think it's possible that the school can end up with 4 lessons instead
of 2.

There's not much in the history to explain whether this is intended or not, so to try to make the test more
deterministic I've decided to run the `destroy_seed_data` task before each expectation. This should reset the system back to a known state so that the expectations are deterministic.

I considered refactoring the task to take the global data as a parameter or stub it out, but I'm worried about breaking something I don't fully understand. Hopefully this will fix the flakiness.


chrislo:
Thanks @chrisroos - let's give it a go.

PR #583, opened 2025-08-25
725-add-import-endpoint

danhalson:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/725

## What's changed?

- Adds import_origin / import_id fields, so we have the ability to match up classes with where it was imported from
- Adds api/classes/import endpoint, with validation on import only to ensure an import origin is defined
- Improves the messaging on ActionDispatch::Http::Parameters::ParseError, so it's clear it's malformed JSON
- Removes some logging from the test_utilities_controller, that looks to be left in by accident

## Steps to perform locally after merging

- `rails db:migrate`

- N/A


app/models/school_class.rb
```diff
@@ -9,13 +9,21 @@ class SchoolClass < ApplicationRecord
 
   scope :with_teachers, ->(user_id) { joins(:teachers).where(teachers: { id: user_id }) }
 
-  before_validation :assign_class_code, on: :create
+  before_validation :assign_class_code, on: %i[create import]
```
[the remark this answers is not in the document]
danhalson: Imported classes are new classes, and won't have an existing class code

app/models/school_class.rb
```diff
@@ -9,13 +9,23 @@ class SchoolClass < ApplicationRecord
...
   validates :code, uniqueness: { scope: :school_id }, presence: true, format: { with: /\d\d-\d\d-\d\d/, allow_nil: false }
   validate :code_cannot_be_changed
   validate :school_class_has_at_least_one_teacher
 
+  # Adding `validate:true` would conflict with `allow_nil: true` (since nil is not an allowable value in the enum)
+  # so we have to add custom validation that only validates non-nil values (`import_origin_must_be_valid_if_present()`)
+  enum :import_origin, { google_classroom: 0 }, allow_nil: true
+
+  validates :import_origin, presence: true, on: :import
+  validate :import_origin_must_be_valid_if_present
```
patch0, line 24: I think this could be simplified to 

```ruby
validates :import_origin, inclusion: { in: import_origins.keys }, if: -> { import_origin.present? }
```
or
```ruby
validates :import_origin, inclusion: { in: import_origins.keys }, allow_nil: true
```
then you don't need the extra validation method

patch0: In fact, [reading the docs]() you should be able to define the enum with that

```ruby
enum :import_origin, { google_classroom: 0 }, validate: { allow_nil: true }
```

patch0 (approved):
Looks good!

PR #585, opened 2025-08-28
853-remove-student-task

danhalson:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/853

## What's changed?

- Adds a service and a task that utilises it to facilitate removing students via either a csv or comma seperated list.

N/A

### DIFF
--- a/lib/tasks/remove_students.rake
+++ b/lib/tasks/remove_students.rake
@@ -33,6 +33,20 @@ namespace :remove_students do
       exit 1
     end
 
+    puts "\n===================="
+    puts "Student removal options summary\n"
+    puts "SCHOOL: #{school.name} (#{school.id})"
+    puts "REMOVE_FROM_PROFILE: #{remove_from_profile}"
+    puts "Students to remove: #{students.size}"
+    puts "====================\n\n"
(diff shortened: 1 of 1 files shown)


lib/tasks/remove_students.rake
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+require_relative '../../app/services/student_removal_service'
+require 'csv'
+
+namespace :remove_students do
```
grega, line 6: I would suggest adding a "dry run" flag to this task so that it can be run, and report on its actions, without affecting any data.

danhalson: haha, I started with one, but when I factored it out into the service that became tricky - and I felt it wouldn't be doing much, it could at least say how many students it would affect?

grega:
Not a comment on the code as such, but am thinking it would be worth prompting anyone running this in future to initiate a manual Heroku Postgres backup on both Profile and Editor API, eg. https://devcenter.heroku.com/articles/heroku-postgres-backups#manual-backups

However that's best captured here! Perhaps as part of a confirmation prompt?

> Please confirm deletion of [x] users, and that recent Postgres backups have been captured for all services affected (https://devcenter.heroku.com/articles/heroku-postgres-backups#manual-backups)

PR #586, opened 2025-09-02
Always respond with JSON from API endpoints

floehopper:
## Status

- Related to https://github.com/RaspberryPiFoundation/experience-cs/issues/1197

## Points for consideration:

- Security: I've only included the exception class and message in the JSON response to avoid accidentally leaking other information.
- Performance: I'm confident this won't have any significant negative impact on performance.

## What's changed?

* Previously, when an unhandled exception was raised within an API controller action, the HTTP response was the static `public/500.html` HTML error page with a 500 status code. Now the HTTP response has a JSON body including a string representation of the error which is less surprising, more consistent, and easier for clients to parse.
* Previously, when an exception was explicitly handled in one of the `rescue_from` statements in `ApiController`, the HTTP response was a HEAD response (i.e. no body) with a specific status code. Now the HTTP response has a JSON body including a string representation of the error which is less surprising, more consistent, and easier for clients to parse. The status code is unchanged.
* Since both of these changes are additive, I'm confident that they won't break `editor-standalone` or `experience-cs`:
    * The code in the files in [`src/utils/apiHandler`](https://github.com/RaspberryPiFoundation/editor-standalone/tree/2a9d342c9f0d378a0b2b5a87c1ec5abc63bb6c1e/src/utils/apiCallHandler) in `editor-standalone` all use the `axios` package in its default configuration, which means that [any status code outside the 2XX range](https://axios-http.com/docs/handling_errors) will be treated as an error and an exception will be raised. Since the changes in this PR don't change the status code, `editor-standalone` should continue to work as before.
    * Similarly the code in [`lib/editor_api/client.rb`](https://github.com/RaspberryPiFoundation/experience-cs/blob/ed3653e902cb10c20de820cd85e99b2411eea94e/lib/editor_api/client.rb) in `experience-cs` uses status codes to detect errors and so shouldn't be affected either.

N/A


fspeirs (commented):
I'm not in a position to approve in the wider context of editor-api, but I've reviewed these changes and they make sense to me.

danhalson (approved):
Thanks this is great 👌

PR #588, opened 2025-09-03
Updates the service to only process a single student at a time and to raise on errors, allowing it to be more adaptable. Updated the task to handle iterating over students and passing to service, and handling errors, allowing better output. More comprehensive tests also added.

danhalson:
## Status

Ready for review

## What's changed?

- Updates the service to only process a single student at a time and to raise on errors, allowing it to be more adaptable. 
- Updated the task to handle iterating over students and passing to service, and handling errors, allowing better output. 
- Adds usage to top of task
- More comprehensive tests also added.

N/A

### DIFF
--- a/app/services/student_removal_service.rb
+++ b/app/services/student_removal_service.rb
@@ -1,42 +1,63 @@
 # frozen_string_literal: true
 
 class StudentRemovalService
-  def initialize(students:, school:, remove_from_profile: false, token: nil)
-    @students = students
+  class NoSchoolError < StandardError; end
+  class NoClassesError < StandardError; end
+  class StudentHasProjectsError < StandardError; end
+  class NoopError < StandardError; end
+
+  def initialize(school:, remove_from_profile: false, token: nil, raise_on_noop: false)
     @school = school
     @remove_from_profile = remove_from_profile
     @token = token
+    @raise_on_noop = raise_on_noop
   end
 
-  # Returns an array of hashes, one per student, with details of what was removed
-  def remove_students
-    results = []
-    @students.each do |user_id|
-      result = { user_id: }
-      begin
-        # Skip if student has projects
-        projects = Project.where(user_id: user_id)
-        result[:skipped] = true if projects.length.positive?
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def remove_student(student_id)
+    raise NoSchoolError, 'School not found' if @school.nil?
+    raise NoClassesError, 'School has no classes' if @school.classes.empty?
+    raise StudentHasProjectsError, 'Student has existing projects' if Project.exists?(user_id: student_id)
 
-        unless result[:skipped]
-          ActiveRecord::Base.transaction do
-            # Remove from classes
-            class_assignments = ClassStudent.where(student_id: user_id)
-            class_assignments.destroy_all
+    ActiveRecord::Base.transaction do
+      roles_destroyed = remove_roles(student_id)
+      classes_destroyed = remove_from_classes(student_id)
+      remove_from_profile(student_id) if should_remove_from_profile?
 
-            # Remove roles
-            roles = Role.student.where(user_id: user_id)
-            roles.destroy_all
-          end
-
-          # Remove from profile if requested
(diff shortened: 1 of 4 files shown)


adrian-rpf (approved):
This all makes sense and some nice tweak improvements. 
Worth looking at the Copilot suggestions some of them make sense.

app/services/student_removal_service.rb
```diff
@@ -1,42 +1,63 @@
...
+    raise NoSchoolError, 'School not found' if @school.nil?
+    raise NoClassesError, 'School has no classes' if @school.classes.empty?
+    raise StudentHasProjectsError, 'Student has existing projects' if Project.exists?(user_id: student_id)
+
+    ActiveRecord::Base.transaction do
+      roles_destroyed = remove_roles(student_id)
+      classes_destroyed = remove_from_classes(student_id)
+      remove_from_profile(student_id) if should_remove_from_profile?
+
+      raise NoopError, 'Student has no roles or class assignments to remove' if roles_destroyed.zero? && classes_destroyed.zero? && should_raise_on_noop?
```
patch0, line 27: Does anyone care if it is a noop?  Seems a bit extra, and would simplify code if it didn't need to check this.

app/services/student_removal_service.rb
```diff
@@ -1,42 +1,63 @@
...
-          ProfileApiClient.delete_school_student(token: @token, school_id: @school.id, student_id: user_id) if @remove_from_profile && @token.present?
-        end
-      rescue StandardError => e
-        result[:error] = "#{e.class}: #{e.message}"
-      end
-      results << result
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def remove_student(student_id)
+    raise NoSchoolError, 'School not found' if @school.nil?
+    raise NoClassesError, 'School has no classes' if @school.classes.empty?
```
patch0, line 19: Is this something that is important to check?  I don't think the caller cares if there are no classes.  They might want the student removing from the school roles, or from profile.

spec/lib/tasks/remove_students_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      allow(Rails.logger).to receive(:error)
+    end
+
+    describe 'envvar validation' do
+      it 'exits when SCHOOL_ID is missing' do
+        expect { task.invoke }.to raise_error(SystemExit)
+      end
+
+      it 'exits when school is not found' do
+        ENV['SCHOOL_ID'] = 'non-existent-id'
```
patch0, line 38: Use [ClimateControl](https://github.com/thoughtbot/climate_control) to manage `ENV` in testing.

app/services/student_removal_service.rb
```diff
@@ -1,42 +1,63 @@
 # frozen_string_literal: true
 
 class StudentRemovalService
-  def initialize(students:, school:, remove_from_profile: false, token: nil)
-    @students = students
+  class NoSchoolError < StandardError; end
+  class NoClassesError < StandardError; end
+  class StudentHasProjectsError < StandardError; end
+  class NoopError < StandardError; end
```
patch0, line 7: If you set up a `BaseError` class, then in the rake task you can just rescue anything
```ruby
  # Set up a BaseError class so we can rescue all errors
  BaseError = Class.new(StandardError)
  ## Specific errors
  NoSchoolError = Class.new(BaseError)
  NoClassesError = Class.new(BaseError)
  StudentHasProjectsError = Class.new(BaseError)
  NoopError = Class.new(BaseError)
```
then in the rake task

```
begin
  ...
rescue StudentRemovalService::BaseError => e
  # whatever it is here
end
```

app/services/student_removal_service.rb
```diff
@@ -1,42 +1,63 @@
...
+
+  def initialize(school:, remove_from_profile: false, token: nil, raise_on_noop: false)
```
patch0, line 9: I guess a fundamental question is why is this asking for a school when the service is a "student removal service"?  I'd suggest maybe renaming it to `SchoolStudentsRemovalService` or let it handle student IDs directly, and pass them in one at a time.

danhalson: but yes, that's good call about altering the name - that'll help make it clearer

adrian-rpf:
@danhalson this PR seems to have gone stale. Any update?

PR #600, opened 2025-10-01
Fix to hopefully resolve the flakey seed test by ensuring different ids are used

danhalson:
## Status

Ready to review

## What's changed?

Fix to resolve the flakey seed test by ensuring different ids are used

N/A

### DIFF
--- a/spec/lib/tasks/for_education_spec.rb
+++ b/spec/lib/tasks/for_education_spec.rb
@@ -3,13 +3,19 @@
 require 'rails_helper'
 require 'rake'
 require 'climate_control'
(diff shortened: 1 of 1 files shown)


spec/lib/tasks/for_education_spec.rb
```diff
@@ -58,8 +65,9 @@
     let(:school) { School.find_by(creator_id:) }
 
     before do
-      Rake::Task['for_education:destroy_seed_data'].invoke
```
loiswells97: Why is this no longer needed? Or was it never needed?

PR #610, opened 2025-11-07
Add ability to set read at on feedback

jamiebenstead:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/992

## What's changed?

- Added read_at datetime to the feedback table
- Created endpoint to set feedback as read
- Added functionality to set a datetime on the read_at field on feedback
- Added tests

### DIFF
--- a/spec/concepts/feedback/set_read_at_spec.rb
+++ b/spec/concepts/feedback/set_read_at_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe Feedback::SetRead, type: :unit do
 
       it 'returns the updated feedback' do
         response = described_class.call(feedback: feedback)
-        expect(response[:feedback]).to eq(feedback)
+        expect(response[:feedback]).to be_a(Feedback)
       end
 
       it 'returns read_at' do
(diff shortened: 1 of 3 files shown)


spec/concepts/feedback/set_read_at_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      it 'returns read_at as a timestamp' do
+        response = described_class.call(feedback: feedback)
+        expect(response[:feedback].read_at).to be_a(ActiveSupport::TimeWithZone)
+      end
+    end
+
+    context 'when set_read fails' do
+      before do
+        allow(feedback).to receive(:save!).and_raise(StandardError, 'Transition failed')
```
loiswells97: Very nitpicky sorry, but there might be a better example of an error message, just because Transition failed is quite specific to the state machine stuff

spec/concepts/feedback/set_read_at_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      it 'returns a failed operation response' do
+        response = described_class.call(feedback: feedback)
+        expect(response.success?).to be(false)
+      end
+
+      it 'returns the original feedback' do
+        response = described_class.call(feedback: feedback)
+        expect(response[:feedback]).to eq(feedback)
+      end
```
loiswells97: this is the same test as the one you have above, where response[:feedback] should have changed, so I'm confused

loiswells97 (approved):
Looks great, let's get it merged! 🎉

PR #637, opened 2025-12-04
Implement Endpoint to allow Mass Student Deletion

fspeirs:
## Status

- Closes RaspberryPiFoundation/experience-cs#1597
- Related to RaspberryPiFoundation/experience-cs#999

## Points for consideration:

- This PR changes the behaviour of `StudentRemovalService` from "skips deleting users that have projects" to "unconditionally deletes users and their projects" (see [Slack thread from 2025-12-03](https://raspberrypifoundation.slack.com/archives/C08831T17H6/p1764761547717129))

## What's changed?

- Adds a `SchoolStudentsController#destroy_batch` endpoint for batch deletion of students, only accessible to school owners (not teachers).
- Modifies the `SchoolStudentsController#destroy` endpoint to wrap `SchoolStudentsController#destroy_batch`, which leads its return code to change from `204` to `200`.

Nothing additional required.

## Tasks Remaining Before Merge

- [x] FS to confirm that there are no safeguarding concerns around mass user- and user-content-deletion. - Phil Howell signed off on this feature in [this document](https://docs.google.com/document/d/1Ecv_ZBqQuItyH7qA57-W8STsIVyXCCnB9FBfsSsgVMA/edit?tab=t.0#heading=h.luara7hxgofz)

### DIFF
--- a/app/controllers/api/school_students_controller.rb
+++ b/app/controllers/api/school_students_controller.rb
@@ -145,7 +145,6 @@ module Api
     private
 
     def remove_students(student_ids)
-      # Invoke StudentRemovalService
       service = StudentRemovalService.new(
         students: student_ids,
         school: @school,
--- a/app/services/student_removal_service.rb
+++ b/app/services/student_removal_service.rb
@@ -13,7 +13,6 @@ class StudentRemovalService
     results = []
 
     @students.each do |user_id|
-      # Ensure that the student has a role in this school and skip if not.
       student_roles = Role.student.where(user_id:, school_id: @school.id)
       if student_roles.empty?
         results << { user_id:, skipped: true, reason: 'no_role_in_school' }
@@ -35,7 +34,6 @@ class StudentRemovalService
           student_roles.destroy_all
 
           # Remove from profile if requested - inside transaction so it can be rolled back
-          # If this call fails, the entire transaction will be rolled back
           ProfileApiClient.delete_school_student(token: @token, school_id: @school.id, student_id: user_id) if @remove_from_profile && @token.present?
         end
       rescue StandardError => e


app/controllers/api/school_students_controller.rb
```diff
@@ -110,17 +110,63 @@ def update
...
+
+      # Remove duplicates to avoid redundant processing
+      unique_student_ids = student_ids.uniq
+      remove_students(unique_student_ids)
     end
 
     private
 
+    def remove_students(student_ids)
+      # Invoke StudentRemovalService
```
zetter-rpf: Not a reason to block merging, but I found some of these comments in this change to be redundant. For example here it's not useful to have a comment saying that you are about to invoke a method.

Comments have a cost to maintain just like code and can become out of date or be misleading.

app/services/student_removal_service.rb
```diff
@@ -11,25 +11,31 @@ def initialize(students:, school:, remove_from_profile: false, token: nil)
...
-            roles = Role.student.where(user_id: user_id)
-            roles.destroy_all
-          end
+          # Remove roles
+          student_roles.destroy_all
 
-          # Remove from profile if requested
+          # Remove from profile if requested - inside transaction so it can be rolled back
+          # If this call fails, the entire transaction will be rolled back
           ProfileApiClient.delete_school_student(token: @token, school_id: @school.id, student_id: user_id) if @remove_from_profile && @token.present?
```
zetter-rpf: Having API calls in transactions can cause problems as they will make the transaction slower and make it more likely to cause contention. I think this is probably fine as I wouldn't expect this endpoint to be receive particularly traffic.

PR #771, opened 2026-04-08
Use headers instead of cookies for authorization

abcampo-iry:
## Status

- Closes [1256](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1256)
- Related to [1239](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1239) and [1255](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1255)

## What's changed?

- Force endpoints to use authentication headers instead of Cookies
- Removed cookie helpers since not used anymore


spec/requests/projects/show_spec.rb
```diff
@@ -54,41 +54,15 @@
...
+    context 'when loading a scratch project' do
+      let!(:project) { create(:project, school:, user_id: teacher.id, locale: nil, project_type: Project::Types::CODE_EDITOR_SCRATCH) }
 
-      it 'does not set auth cookie when project is not scratch' do
+      it 'does not set a scratch auth cookie' do
         get("/api/projects/#{project.identifier}", headers:)
 
         expect(response).to have_http_status(:ok)
         expect(response.cookies['scratch_auth']).to be_nil
       end
```
zetter-rpf: I think this test doesn't have any purpose now - I think it's very unlikely that we would unintentionally set this cookie on this action in the future

zetter-rpf (approved):
Great! 

One very optional suggestion.

spec/requests/projects/show_spec.rb
abcampo-iry: I agree.
Moreover, my thoughts was to have a negative test and then remove the test completely after the PR is completed.
Happy to remove it in the same PR.

At least with this comment we leave record that it passed and we can remove it.

PR #783, opened 2026-04-14
Issues/1263 scope assets to project in scratch

abcampo-iry:
## Status

- Closes partially [1256](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1263)

## What’s changed?

Scratch assets are now scoped by project and uploader instead of being treated as globally shared.
- scratch_assets now stores:
- project_id
- uploaded_user_id

That changes ownership like this:

- global Scratch library assets remain global: project_id = nil, uploaded_user_id = nil
- project-scoped assets are unique per (project_id, uploaded_user_id, filename)
- different projects can upload the same asset filename
- different users on the same teacher project can upload the same filename without colliding

### Scratch asset reads and uploads

The Scratch asset endpoints now behave like this:
- X-Project-ID is required
- reads and uploads are authorized against that active Scratch project
- uploads always write to the active project from the header, with uploaded_user_id = current_user.id
- asset endpoints no longer create or reuse remixes

Asset lookup now resolves in this order:
- the current user’s asset on the active project
- the active project owner’s/shared asset
- the current user’s asset on ancestor projects
- ancestor owner/shared assets
- the global library asset

This means:
- a project can load its own assets
- remixes can still load assets from ancestor projects
- Scratch library assets still work as global assets

For SVGs:
- project SVG assets are served as application/octet-stream
- only global library SVG assets are served as image/svg+xml

### Scratch remix behavior

Remix creation still reuses an existing remix for the same user and original project instead of creating duplicates.
The tricky case is when a student adds a Scratch asset before pressing Save for the first time:

- the asset is stored on the original project, but with uploaded_user_id = student.id
- that makes it visible only to that student while they are still working from the original project
- when the student does the first explicit remix save, the remix is created or reused
- after that, those uploaded assets are reassigned from the original project to the remix
- later saves keep using that same remix

This keeps remix logic inside the remix flow, instead of mixing remix creation into the asset endpoints.

### DIFF
--- a/app/controllers/api/scratch/projects_controller.rb
+++ b/app/controllers/api/scratch/projects_controller.rb
@@ -110,8 +110,6 @@ module Api
             uploaded_user_id:
           )
         end
-      rescue StandardError => e
-        Sentry.capture_exception(e)
       end
 
       def skip_scratch_asset_reassignment?(original_project:, remix_project:, uploaded_user_id:)
--- a/app/controllers/api/scratch/projects_controller.rb
+++ b/app/controllers/api/scratch/projects_controller.rb
@@ -12,7 +12,7 @@ module Api
       before_action :ensure_create_is_a_remix, only: %i[create]
 
       def show
-        render json: scratch_project_content(@project.scratch_component.content.to_h)
+        render json: scratch_project_content_with_stage_first(@project.scratch_component.content.to_h)
       end
 
       def create
@@ -82,10 +82,11 @@ module Api
         params.slice(:meta, :targets, :monitors, :extensions).to_unsafe_h
       end
 
-      def scratch_project_content(content)
+      def scratch_project_content_with_stage_first(content)
         targets = content['targets']
         return content unless targets.is_a?(Array)
 
+        # Scratch expects the stage target first and can fail to load otherwise.
         stage_targets, other_targets = targets.partition do |target|
           target.is_a?(Hash) && (target['isStage'] || target[:isStage])
         end
--- a/app/controllers/api/scratch/projects_controller.rb
+++ b/app/controllers/api/scratch/projects_controller.rb
@@ -27,7 +27,7 @@ module Api
           scratch_component = existing_remix.scratch_component || existing_remix.build_scratch_component
           scratch_component.content = scratch_content_params
           existing_remix.save!
-          reassign_uploaded_scratch_assets(original_project:, remix_project: existing_remix)
+          move_assets_uploaded_by_current_user_before_remix(original_project:, remix_project: existing_remix)
 
           return render json: { status: 'ok', 'content-name': existing_remix.identifier }, status: :ok
         end
@@ -42,7 +42,7 @@ module Api
         )
 
         if result.success?
-          reassign_uploaded_scratch_assets(original_project:, remix_project: result[:project])
+          move_assets_uploaded_by_current_user_before_remix(original_project:, remix_project: result[:project])
           render json: { status: 'ok', 'content-name': result[:project].identifier }, status: :ok
         else
           render json: { error: result[:error] }, status: :bad_request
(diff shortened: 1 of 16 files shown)


app/controllers/api/scratch/projects_controller.rb
```diff
@@ -68,6 +81,55 @@ def load_original_project(identifier)
...
+        return content unless targets.is_a?(Array)
+
+        # Scratch expects the stage target first and can fail to load otherwise.
+        stage_targets, other_targets = targets.partition do |target|
+          target.is_a?(Hash) && (target['isStage'] || target[:isStage])
+        end
+        return content if stage_targets.empty?
+
+        content.merge('targets' => stage_targets + other_targets)
+      end
```
abcampo-iry: I found this issue when i was trying to save:
For SB3 project JSON, targets[0] is expected to be the Stage.
The official Scratch parser schema models it like this:
```json
"targets": {
    "type": "array",
    "items": [
      { "allOf": [stage, target] }
    ],
    "additionalItems": {
      "allOf": [sprite, target]
  }
}
```

app/controllers/api/scratch/projects_controller.rb
```diff
@@ -68,6 +81,54 @@ def load_original_project(identifier)
...
+
+        ScratchAsset.where(project: original_project, uploaded_user_id:).find_each do |source_asset|
+          reassign_uploaded_scratch_asset(
+            source_asset:,
+            remix_project:,
+            uploaded_user_id:
+          )
+        end
+      rescue StandardError => e
+        Sentry.capture_exception(e)
```
zetter-rpf: I know this is a pattern already in the codebase but I would like to remove it (and not introduce any more instances of it).

By rescuing all errors it masks real problems especially in development and test where sentry errors are less visible.

If we don't expect any errors, we should remove it. If we do expect errors and want to them be silenced, we should rescue that specific error (and if it's expected we should consider not sending it to sentry)

app/controllers/api/scratch/projects_controller.rb
```diff
@@ -68,6 +81,54 @@ def load_original_project(identifier)
...
+
+      def reassign_uploaded_scratch_assets(original_project:, remix_project:)
+        uploaded_user_id = current_user.id
+        return if skip_scratch_asset_reassignment?(
+          original_project:,
+          remix_project:,
+          uploaded_user_id:
+        )
+
+        ScratchAsset.where(project: original_project, uploaded_user_id:).find_each do |source_asset|
```
zetter-rpf: Is this right? It looks like it will only re-assign the asset if it was uploaded by the current user. 
When remixing it's unlikely this will be the case - normally a teacher would upload the asset and a student would remix it.

app/controllers/api/scratch/projects_controller.rb
```diff
@@ -68,6 +81,54 @@ def load_original_project(identifier)
       def scratch_content_params
         params.slice(:meta, :targets, :monitors, :extensions).to_unsafe_h
       end
+
+      def scratch_project_content(content)
```
zetter-rpf: This might be easier to test if it was a class extracted from the controller.

app/controllers/api/scratch/projects_controller.rb
```diff
@@ -68,6 +81,54 @@ def load_original_project(identifier)
...
+
+        stage_targets, other_targets = targets.partition do |target|
+          target.is_a?(Hash) && (target['isStage'] || target[:isStage])
+        end
+        return content if stage_targets.empty?
+
+        content.merge('targets' => stage_targets + other_targets)
+      end
+
+      def reassign_uploaded_scratch_assets(original_project:, remix_project:)
```
zetter-rpf: This might also be clearer and easier to test if it was a class outside of the controller

app/controllers/api/scratch/projects_controller.rb
abcampo-iry: Let me remove this

app/controllers/api/scratch/projects_controller.rb
abcampo-iry: I'm happy to move some of those to its own object, i did it initially, but I wasn't sure if we would prefer to keep it as part of the model or as its own.
I do like it there as well, so for me it's 50/50

app/models/scratch_asset.rb
```diff
@@ -1,7 +1,47 @@
...
+    assets_by_project_id = where(filename:, project_id: lineage_projects.map(&:id)).group_by(&:project_id)
+
+    lineage_projects.each do |lineage_project|
+      project_assets_by_user_id = assets_by_project_id.fetch(lineage_project.id, []).index_by(&:uploaded_user_id)
+      visible_asset = project_assets_by_user_id[current_user_id] || project_assets_by_user_id[lineage_project.user_id]
+      return find(visible_asset.id) if visible_asset
+    end
+
+    global_assets.find_by(filename:)
+  end
```
zetter-rpf, line 26: This looks complex and might benefit from some unit tests

app/controllers/api/scratch/projects_controller.rb
zetter-rpf: It's up to you, I don't feel strongly

app/controllers/api/scratch/projects_controller.rb
zetter-rpf: I've remembered what the purpose of this is now after our conversation last week - I think it's to move assets owned by the current user that were uploaded before the remix started. I wonder if there's a way to make that clearer in naming

zetter-rpf (approved):
This is great, I've made a few optional suggestions that could make this clearer or improve testing.

There is one point in the issue that I don't think this does: 
> Only trusted global assets from Scratch should be served with the image/svg+xml from the ScratchAsset Controller

Maybe it would be best to split that out into a different issue so we can merge this in sooner?

app/controllers/api/scratch/projects_controller.rb
abcampo-iry: Revoved rescue

app/controllers/api/scratch/projects_controller.rb
abcampo-iry: Using a more explicit naming: `move_assets_uploaded_by_current_user_before_remix`

abcampo-iry:
>  There is one point in the issue that I don't think this does:
Only trusted global assets from Scratch should be served with the image/svg+xml from the ScratchAsset Controller

```ruby
def response_content_type
    extension_is_svg = File.extname(filename).casecmp('.svg').zero?
    return 'image/svg+xml' if global? && extension_is_svg
    return 'application/octet-stream' if extension_is_svg

    Marcel::MimeType.for(name: filename) || 'application/octet-stream'
  end
```

@zetter-rpf 
I was trying to cover this part in here, using the `global?`. I will merge, but if you think is not finished we can open another issue.

zetter-rpf:
> > There is one point in the issue that I don't think this does:
> > Only trusted global assets from Scratch should be served with the image/svg+xml from the ScratchAsset Controller
> 
> ```ruby
> def response_content_type
>     extension_is_svg = File.extname(filename).casecmp('.svg').zero?
>     return 'image/svg+xml' if global? && extension_is_svg
>     return 'application/octet-stream' if extension_is_svg
> 
>     Marcel::MimeType.for(name: filename) || 'application/octet-stream'
>   end
> ```
> 
> @zetter-rpf I was trying to cover this part in here, using the `global?`. I will merge, but if you think is not finished we can open another issue.

That's great then, sorry for missing that.

PR #784, opened 2026-04-15
Upgrade to Rails 8.1

zetter-rpf:
Upgrade to Rails 8.1

This is so we get the latest features and bugfixes.

Rails 8 release notes: https://guides.rubyonrails.org/8_0_release_notes.html
Rails 8.1 release notes: https://guides.rubyonrails.org/8_1_release_notes.html

Some notable changes this deals with:

+ Introduces `expect` as an alternative to `permit` to prevent clients being able to trigger 500s. Note that this isn't a mandatory change, but rubocop now warns about uses of permit so I have updated them - (more here)[https://blog.saeloun.com/2024/12/10/rails-8-adds-parametersexpect-to-safely-filter-and-require-params/]
+ columns in schema.rb are now ordered alphabetically. This prevents 'competing schemas' when two people add the columns to the same table at similar times.


zetter-rpf:
This is mostly done, there are some to_time warnings that I wanted to review (possibly fixed by enabling the setting in the upgrade file)

app/controllers/api/projects/remixes_controller.rb
```diff
@@ -67,17 +67,19 @@ def load_and_authorize_remix_identifier
...
+                            :identifier,
+                            :project_type,
+                            :locale,
+                            :user_id,
+                            :videos,
+                            :audio,
+                            :instructions,
+                            {
+                              image_list: [],
+                              components: [%i[id name extension content index]]
```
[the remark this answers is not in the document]
zetter-rpf: > In order to permit an array of parameters, the array must be defined explicitly. Use double array brackets, an array inside an array, to declare that an array of parameters is expected.

config/initializers/new_framework_defaults_8_0.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Be sure to restart your server when you modify this file.
+#
+# This file eases your Rails 8.0 framework defaults upgrade.
+#
+# Uncomment each configuration one by one to switch to the new default.
```
abcampo-iry, line 7: Question how, i'm not sure if this file needs to remain, or we should update this in the same PR:

```ruby
module App
  class Application < Rails::Application
    # Initialize configuration defaults for originally generated Rails version.
    config.load_defaults 7.1
```

The recommendations says:
> You should also first upgrade to Rails 8.0 in case you haven't and make sure your application still runs as expected before attempting an update to Rails 8.1

Are we happy with the risk, instead of following the steps 7.1 -> 7.2, 7.2 -> 8.0, 8.0 -> 8.1

app/controllers/api/google_auth_controller.rb
```diff
@@ -48,7 +48,7 @@ def response_error_message
     def google_token_params
       params.require(:google_auth).require(:code)
       params.require(:google_auth).require(:redirect_uri)
-      params.require(:google_auth).permit(:code, :redirect_uri)
+      params.expect(google_auth: %i[code redirect_uri])
```
abcampo-iry, line 51: I was investigating a little bit, small nitpick related to the other changes, I think the syntax needs to be something like this the important part is the expect before permit:

```ruby
  def google_token_params
    params
      .expect(google_auth: %i[code redirect_uri])
      .tap { |google_auth| google_auth.require(%i[code redirect_uri]) }
  end
```

or
```ruby
  def google_token_params
    google_auth_params = params.expect(google_auth: %i[code redirect_uri])
    google_auth_params.require(%i[code redirect_uri])
    google_auth_params
  end
```

abcampo-iry (commented):
I think overall really good, good job, I'm just leaving a couple of comments.

I think my understanding might be wrong but just want to raise it since we might need to do a bit more for the upgrade or maybe do smaller upgrades, so I would love to verify.

config/initializers/new_framework_defaults_8_0.rb
zetter-rpf: Thanks, I missed this - I didn't realise that there were a lot of new defaults in 7.2 and 8.1 and I missed updating the framework defaults (but did run the tests against 8.0)

I think your right, it would be safest to ship the change separately:
+ Update to 7.2 framework defaults
+ Upgrade to rails 8.0 and update framework defaults (with any exceptions we need)
+ Upgrade to rails 8.1 and framework defaults (with any exceptions we need)

app/controllers/api/google_auth_controller.rb
zetter-rpf: Good spot

PR #787, opened 2026-04-21
Store school email domains

PetarSimonovic:
## Status

Closes [#1325](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1325)

## What's changed?

- Add a `school_email_domains` table
- Add a `SchoolEmailDomain` model
- Add a `SchoolEmailDomainValidator` that normalises and checks domains
- Update association on `School`
- Add `valid_domain?` method on `School`
- Add and update specs


spec/models/school_email_domain_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SchoolEmailDomain do
+  let(:school) { create(:school, creator_id: SecureRandom.uuid) }
+
+  it { is_expected.to belong_to(:school) }
+  it { is_expected.to validate_presence_of(:domain) }
+
```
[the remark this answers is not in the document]
PetarSimonovic: Added specs to confirm uniqueness is scoped to school id

app/models/school_email_domain.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  before_validation :format_domain
+
+  private
+
+  def format_domain
+    return if domain.nil?
+
+    self.domain = domain.to_s.strip.sub(/\A@+/, '').downcase
+  end
```
PetarSimonovic: I've applied only minimal formatting to standardise domains - happy to add more if needed

app/models/school_email_domain.rb
```diff
@@ -0,0 +1,18 @@
...
+  validates :domain, uniqueness: { scope: :school_id }
+
+  before_validation :format_domain
+
+  private
+
+  def format_domain
+    return if domain.nil?
+
+    self.domain = domain.to_s.strip.sub(/\A@+/, '').downcase
```
fspeirs: **Issue**: I am not convinced this is a strict-enough validation check for an application like this, for a couple of reasons:

1. It only anticipates one type of error: that the user includes an @ before the domain.
2. It does not check that what has been entered is a valid domain.

Instead, can we consider introducing the [`public_suffix` gem](https://rubygems.org/gems/public_suffix) to get ourselves a stronger guarantee that what is being stored is a valid domain?

spec/models/school_email_domain_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SchoolEmailDomain do
```
fspeirs, line 5: **Suggestion**: If we adopt `public_suffix` we will want different kinds of tests here.

fspeirs (changes requested):
Overall this looks like the right approach. I have one consideration about validating domains - I feel like it would be helpful to be stricter in checking that the user hasn't typo'ed their domain or misunderstood what we are asking for.

app/models/school_email_domain.rb
PetarSimonovic: @fspeirs I've implemented this but please let me know if I've misunderstood or if there's any further validation required

app/models/school_email_domain.rb
```diff
@@ -0,0 +1,43 @@
...
+  before_validation :normalise_domain
+  validate :validate_public_suffix
+
+  private
+
+  def normalise_domain
+    return if domain.nil?
+
+    self.domain = build_normalised_domain_string(domain)
+  end
```
fspeirs: **Issue:** I think this code could be much simpler and doesn't have to expressly handle specific things like `uri_host_if_http_url`. Instead, I suggest that we just try to parse a URI and, if successful take the host portion:

```ruby
def normalise_domain
  return if domain.blank?
  value = domain.strip.downcase
  # Add a scheme unless it already has one, so URI can parse it
  value = "http://#{value}" unless value =~ %r{\A[a-z][a-z0-9+\-.]*://}i
  
  uri = URI.parse(value)

  host = uri.host
  return if host.blank?

  if PublicSuffix.valid?(host)
    self.domain = host
  else
    errors.add(:domain, "is not a valid domain")
  end
rescue URI::InvalidURIError
  errors.add(:domain, "is not a valid domain")
end
```

This is kind of the approach you're taking in `uri_host_if_http_url` and pulling it up to be the main way we normalise domains: (a) can we parse it with `URI::parse` and secondly is it a valid suffix.

app/models/school.rb
```diff
@@ -6,6 +6,7 @@ class School < ApplicationRecord
   has_many :projects, dependent: :nullify
   has_many :roles, dependent: :nullify
   has_many :school_projects, dependent: :nullify
+  has_many :school_email_domains, dependent: :destroy
```
fspeirs, line 9: **Issue:** Callers (including my work) would like a validator method on this called something like: `valid_domain?(candidate_domain)` which returns true if `candidate_domain` is registered for this school.

This would allow testing domains presented by registering users without exposing the internal implementation of `school_email_domains`.

fspeirs (changes requested):
Directionally, quite correct. Just some sharpening of the normalisation approach in `school_email_domain` and a request for a testing method in `school.rb`. The lack of a testing method is non-blocking if it's coming in a separate PR.

peconia:
The regex we use in profile side to validate is ```'^\\s*(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\\.)+[A-Za-z]{2,63}\\s*$'```

While it is similar, we would reject things that had a dash in the wrong place, like `example.-wrong.edu`, but that would be approved by this one. I also wonder about things that are domains but not email domains, like having `www.` or `ftp.` prefix to the domain. Maybe that will come down to making sure the UI is clear in what is expected, and the API level can validate some things too.

spec/models/school_email_domain_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+  context 'with an invalid domain' do
+    it { is_expected.not_to allow_value('').for(:domain) }
+    it { is_expected.not_to allow_value('   ').for(:domain) }
+    it { is_expected.not_to allow_value('http://').for(:domain) }
+    it { is_expected.not_to allow_value('edu').for(:domain) }
+    it { is_expected.not_to allow_value('com').for(:domain) }
+    it { is_expected.not_to allow_value('co.uk').for(:domain) }
+    it { is_expected.not_to allow_value('http://invalid uri').for(:domain) }
+    it { is_expected.not_to allow_value('-wrong.edu').for(:domain) }
```
PetarSimonovic, line 87: @fspeirs and @peconia I've grouped the invalid domains together so it might be easier to check edge cases like the one that Päivi mentioned

app/models/school_email_domain.rb
```diff
@@ -0,0 +1,42 @@
...
+    host = uri.host&.delete_suffix('.')
+
+    validate_host(host)
+  rescue URI::InvalidURIError
+    errors.add(:domain, :invalid)
+  end
+
+  def validate_host(host)
+    accounts_host_format =
+      /\A\s*(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\.)+[A-Za-z]{2,63}\s*\z/i
```
PetarSimonovic: @peconia we're validating against the regex that Profile uses

app/models/school_email_domain.rb
PetarSimonovic: Thanks for this Fraser.

I've implemented your changes and taken account of Päivi's comment below

app/models/school.rb
PetarSimonovic: I've added a validator method to `School` using a straightforward `school_email_domains.exists?` 

I don't think we need to optimise for batch domain checks, since we're validating one student at a time

If we ever did need to perform some sort of batch validation, we could revisit this and perhaps pluck then memoise the domains

zetter-rpf (approved):
Nice one

fspeirs (approved):

PR #793, opened 2026-04-27
Setup vscode tasks for rspec

zetter-rpf:
Set up tasks to make it easier to run rspec, similar to https://github.com/RaspberryPiFoundation/editor-ui/pull/1344 and https://github.com/RaspberryPiFoundation/editor-standalone/pull/780

Note that you will need to have asdf on your non-interactive shell path for this to work. If you see a 'bundler not installed' message this is likely the cause.

You can fix this by making sure asdf is added to your path in your `.zprofile` instead of your `.zshrc`.

I've added the rspec bin stub as part of this as it's quicker than typing bundle exec and simpler than using it in the tasks.json


mwtrew (approved):
I don't use VS Code so can't test this properly, but LGTM and I think the best way to test it is to get everyone using it.

PR #797, opened 2026-04-29
Add batching parameter to salesforce_sync:* tasks

fspeirs:
In preparation for completing the back-fill of School data into Salesforce, this commit adds a LIMIT env var to the salesforce_sync:* rake tasks to avoid overwhelming Salesforce when we sync the back-fill.

The `LIMIT` env var can be set to any value, or `0` for no limit. Absent, it defaults to enqueueing 25 jobs.

## What's changed?

- Wrapped each task in a simple counter that checks for a `Salesforce::` record and, if not found, enqueues a task.
- Aborts when `count` reaches `LIMIT`.

To complete the back-fill:

```bash
LIMIT=10 bundle exec rake salesforce_sync:{school, role, contact}
```

### DIFF
--- a/lib/tasks/salesforce_sync.rake
+++ b/lib/tasks/salesforce_sync.rake
@@ -1,24 +1,75 @@
 # frozen_string_literal: true
 
 namespace :salesforce_sync do
+  # Sync Schools to Salesforce that don't already exist there.
+  # Usage:
+  #   rake salesforce_sync:school          # Default: sync up to 25 records
+  #   LIMIT=100 rake salesforce_sync:school # Sync up to 100 records
+  #   LIMIT=0 rake salesforce_sync:school   # Sync all missing records (unlimited)
   desc 'Sync all Schools to Salesforce'
   task school: :environment do
+    limit = ENV['LIMIT'] ? ENV['LIMIT'].to_i : 25
+    limit = nil if limit.zero?
+    enqueued = 0
+    checked = 0
+
     School.find_each do |school|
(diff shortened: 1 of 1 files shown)


zetter-rpf:
A bit more context in the PR description or a linked issue would help me review and would be useful for people looking at the code in the future. My main questions are:

- Why are we worried about overwhelming salesforce? Is there a rate limit or error we're trying to avoid? 
- How are you intending to run this in production- are you going to run the rake tasks repeatedly until everything is synced?

fspeirs:
@zetter-rpf It was a request from the Salesforce team. I believe there are a number of post-ingest steps that may be run after import which may exhaust account credits capacity if too many arrive at once.

As for running in production, the intention was to probably run two or three times with `LIMIT=200` until all are back-filled into Salesforce.

zetter-rpf:
> As for running in production, the intention was to probably run two or three times with LIMIT=200 until all are back-filled into Salesforce.

How do you know how big a batch to run or how frequently to run them?

Without knowing that I can't judge wether this solves the problem. And I imagine this would be very useful to know for someone running this in the future (even if the answer is just ask the Salesforce team).

fspeirs:
I'm putting this into draft until I can get a bit more clarity from the Salesroce team.

kylec-rpf:
Pushing this to next sprint. We have higher priority work to look at first.

fspeirs:
Dropping this PR without closing it as we have resolved concerns with the Salesforce team and this is no longer needed.

PR #799, opened 2026-04-29
Use Rails 7.2 defaults

zetter-rpf:
Split out of #784 

You can see new settings at [1], they all seem safe for us to switch to.

This is in preparation for upgrading to Rails 8 and later using those defaults.

I'll monitor RAM usage in Heroku after deploying because of the YJIT change.

[1] https://redirect.github.com/rails/rails/blob/v7.2.3.1/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_7_2.rb.tt

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,7 @@ gem 'propshaft'
 gem 'puma', '~> 7.2'
 gem 'rack_content_type_default', '~> 1.1'
 gem 'rack-cors'
-gem 'rails', '~> 7.1'
+gem 'rails', '~> 7.2'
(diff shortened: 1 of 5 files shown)


abcampo-iry (approved):
I just checked, we have in Gemfile:
gem 'rails', '~> 7.1'
and in Lockfile 
7.2.3.1

maybe we want to make it explicit, in the upgrade by adding in Gemfile:
gem 'rails', '~> 7.2'

(although we are effecitvely in that version)

otherwise good, thanks

PR #800, opened 2026-04-29
1272: wire up form to pardot - explore using form handler

DNR500:
issue: [1272](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1272)

# Form Handler for Subscription form

## Overview
This PR explores using a Pardot Form Handler on the backend to send subscription data via `editor-api`.

---

## Key Considerations (Form Handler)

- **Non-standard handling required for Form Handler integration**
Because Pardot Form Handlers do not provide a standard, structured API response contract, editor-api has had to use heuristic response handling (status/body pattern checks). This means outcomes are inferred rather than definitively confirmed on every request, and frontend error detail is necessarily less precise than a typical JSON API integration.

---

### Request

- **Method:** POST  
- **URL:** http://localhost:3009/api/subscriptions  

**Headers:**
- Content-Type: application/json  
- Accept: application/json  

**Body (raw JSON):**
```json
{
  "subscription": {
    "email": "teacher@example.com",
    "test_opt_in": true,
    "privacy_policy": true
  }
}
```

### Expected Responses

- `200 OK` → valid request, provider accepted  
- `422 Unprocessable Content` → validation error  
- `503 Service Unavailable` → provider unavailable / not configured  
- `502 Bad Gateway` → provider response rejected or ambiguous 

### Invalid Example

```json
{
  "subscription": {
    "email": "bad-email",
    "test_opt_in": true,
    "privacy_policy": false
  }
}
```


app/controllers/api/subscriptions_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+module Api
+  class SubscriptionsController < ApiController
+    def create
+      payload = subscription_params.to_h
+      errors = validation_errors_for(payload)
+
+      if errors.empty?
+        submit_result = subscriptions_submitter.call(form_payload: payload)
+        if submit_result.success?
+          Rails.logger.info('[subscriptions#create] outcome=success')
```
zetter-rpf, line 12: Maybe this logging and the `warn` call bellow aren't needed?

We will already log the request (the params, the response code) so i'm not sure if this is telling anything else useful?

app/controllers/api/subscriptions_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+      )
+    end
+
+    def validation_errors_for(payload)
+      errors = []
+      errors << 'email is required' if payload['email'].blank?
+      errors << 'email is invalid' if payload['email'].present? && !valid_email?(payload['email'])
+      errors << 'privacy_policy must be true' unless payload['privacy_policy'] == true
+      errors
+    end
```
zetter-rpf, line 58: Are these strings going to be used in the frontend? I'm just thinking how we would localize them.

I'm not sure what patterns we tend to use in other places, but we could either send back keys that are translated in the frontend, or use the rails validation system (which has default translations, and can be customized).

zetter-rpf (approved):
This is great, just have a couple of minor questions/suggestions

app/controllers/api/subscriptions_controller.rb
DNR500: I can remove these

app/controllers/api/subscriptions_controller.rb
DNR500: I can remove these

app/controllers/api/subscriptions_controller.rb
DNR500: These aren't sent to the frontend anymore

PR #804, opened 2026-05-01
Reduce N+1s by caching submitted project count (Part 1 of 2)

zetter-rpf:
## Status

- Related to https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1313
- Replaces https://github.com/RaspberryPiFoundation/editor-api/pull/794
- Precedes #805 

## What's changed?

There's currently N+1 queries on both the lessons and classes controller as they both try to show the number of submitted projects.

Originally I tired to solve this by using 'include' to load the correct records, however I don't think this is a very scalable solutions as it means loading every submitted project object memory for classes or lessons (which there may be 1000s, especially when listing classes)

Instead use the pattern of a counter cache. Each time a project transitions to/from submitted, we recalculate the number.

This is the first part of two. I've broken up the work so we can deploy this and check that the counts are correct. If we deployed this in one go there would be a chance the counts are inaccurate if a project transition happened during the deploy.

### After deploy

- [x]  Run `rails lessons:backfill_submitted_projects_count` and verify the results by checking a few lessons and comparing it to `lesson#submitted_count`

### DIFF
--- a/lib/tasks/lessons.rake
+++ b/lib/tasks/lessons.rake
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+namespace :lessons do
+  desc 'Backfill cached submitted project counts for lessons'
+  task backfill_submitted_projects_count: :environment do
+    Lesson.connection.execute <<~SQL.squish
+      WITH submitted_counts AS (
+        SELECT lesson_projects.lesson_id, COUNT(*) AS submitted_projects_count
+        FROM projects lesson_projects
+        INNER JOIN projects remixes ON remixes.remixed_from_id = lesson_projects.id
+        INNER JOIN school_projects ON school_projects.project_id = remixes.id
+        INNER JOIN school_project_transitions ON school_project_transitions.school_project_id = school_projects.id
+        WHERE school_project_transitions.most_recent = TRUE
+          AND school_project_transitions.to_state = 'submitted'
+          AND lesson_projects.lesson_id IS NOT NULL
+        GROUP BY lesson_projects.lesson_id
+      )
+      UPDATE lessons
+      SET submitted_projects_count = COALESCE(submitted_counts.submitted_projects_count, 0)
+      FROM lessons target_lessons
+      LEFT JOIN submitted_counts ON submitted_counts.lesson_id = target_lessons.id
+      WHERE lessons.id = target_lessons.id
+    SQL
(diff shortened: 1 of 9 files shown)


lib/tasks/lessons.rake
```diff
@@ -0,0 +1,25 @@
...
+      WITH submitted_counts AS (
+        SELECT lesson_projects.lesson_id, COUNT(*) AS submitted_projects_count
+        FROM projects lesson_projects
+        INNER JOIN projects remixes ON remixes.remixed_from_id = lesson_projects.id
+        INNER JOIN school_projects ON school_projects.project_id = remixes.id
+        INNER JOIN school_project_transitions ON school_project_transitions.school_project_id = school_projects.id
+        WHERE school_project_transitions.most_recent = TRUE
+          AND school_project_transitions.to_state = 'submitted'
+          AND lesson_projects.lesson_id IS NOT NULL
+        GROUP BY lesson_projects.lesson_id
```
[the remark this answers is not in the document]
zetter-rpf: I've checked that lessons are linked to a single project.

lib/tasks/lessons.rake
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+namespace :lessons do
+  desc 'Backfill cached submitted project counts for lessons'
+  task backfill_submitted_projects_count: :environment do
+    Lesson.connection.execute <<~SQL.squish
+      WITH submitted_counts AS (
+        SELECT lesson_projects.lesson_id, COUNT(*) AS submitted_projects_count
```
mwtrew, line 8: I would have thought that this needs to do the same thing as `recalculate_lesson_submitted_projects_count!`, across all projects. Are we writing our own query for efficiency, or is it something else that I'm missing?

mwtrew (approved):

lib/tasks/lessons.rake
zetter-rpf: Yes I defaulted to doing it this way for efficiency without even checking how many lessons there are.

I've checked and there are `28114` - not loads, but thats enough that updating them all one-after-the-other could create database contention and slow down the site for the couple of minutes it's running.

mwtrew: Nice, thanks :)

PR #810, opened 2026-05-06
Add turnstile token to subscription

cocomarine:
Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1353
(frontend PR: https://github.com/RaspberryPiFoundation/editor-standalone/pull/859)

### Notes
- Generated turnstile secret keys for staging and prod in Cloudflare and added them as `CLOUDFLARE_TURNSTILE_SECRET_KEY` to config var using terraform
- Updated .env.example to reflect this


zetter-rpf (approved):
This is great

My only suggestion is that if the turnstile checking was in it's own class rather than a method in the controller it might be easier to test independently of the controller (and then stub in the controller tests). I saw you used shared examples in the spec to help remove some of this duplication which is good, but can make the tests files harder to work with. Up to you if you think this change is worth it.

zetter-rpf (approved):

Issue #813, opened 2026-05-07
Upgrade critical dependencies


grega:
This application is running software that is considered to be out of date (eg. no security support):

- Rails: should be >=8.1

See: https://digital-ops-dashboard.rpf-internal.org/dependency-report

zetter-rpf:
We're now on 8.1! 📈

PR #818, opened 2026-05-13
Rails 8.1.3

abcampo-iry:
## Status

- Closes [1291](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1291) #813 

## Points for consideration:

https://guides.rubyonrails.org/8_1_release_notes.html

## What's changed?

Only updated the Gemfile and agents.md


AGENTS.md
```diff
@@ -1,5 +1,5 @@
 # Project Overview
-- Rails 7.1 monolith for the Raspberry Pi Code Editor API (REST + GraphQL), served at `editor-api.raspberrypi.org`.
+- Rails 8.1 monolith for the Raspberry Pi Code Editor API (REST + GraphQL), served at `editor-api.raspberrypi.org`.
```
zetter-rpf, line 2: I think having the rails version here is a bit unnecessary

zetter-rpf (approved):
Great! Excited to be on the latest version again.

As before, I think it would be good keep an eye on alerts and metrics when we deploy this - let me know when you want to deploy to production.

AGENTS.md
abcampo-iry: I agree, we can remove this later.

PR #819, opened 2026-05-15
Upgrade to Ruby 4

zetter-rpf:
## What's changed?

Update to the latest version of Ruby (4.0.5) so we get the latest features, performance improvements and bugfixes

This looks there are no breaking changes that affect us, I will test a bit more on staging before deploying to production.

Ruby 4 changes: https://github.com/ruby/ruby/blob/v4.0.0/NEWS.md


abcampo-iry (approved):
It looks good to me

PR #821, opened 2026-05-18
Only handle requests via HOST_URL

### LINKED ISSUE #533: Stop serving apps on herokuapp.com subdomains
https://editor-api-staging.herokuapp.com and https://editor-api-production.herokuapp.com should not be accessible, but currently are.

The application can be configured (via `config.hosts`) to only respond to requests on raspberrypi.org subdomains, eg: https://github.com/RaspberryPiFoundation/experience-cs/blob/main/config/environments/production.rb#L78-L79

mwtrew:
## Status

- Closes https://github.com/RaspberryPiFoundation/editor-api/issues/533

## Points for consideration:

This does not make any allowances for preview deployments, however preview deployments are currently not working for other reasons. Once they are, we should add them to the allow-list.

## What's changed?

- Uses the `HOST_URL` value for `config.hosts` in production configuration, which means the server will reject requests where the `Host` header doesn't match `HOST_URL`. This will prevent the API being accessed via `editor-xyz.herokuapp.com` domains.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -86,10 +86,10 @@ Rails.application.configure do
   config.active_record.dump_schema_after_migration = false
 
   # Enable DNS rebinding protection and other `Host` header attacks.
-  # config.hosts = [
-  #   "example.com",     # Allow requests from example.com
-  #   /.*\.example\.com/ # Allow requests from subdomains like `www.example.com`
-  # ]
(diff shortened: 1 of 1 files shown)


zetter-rpf (approved):
👍

mwtrew:
Thanks - will merge this when I have time to test it afterwards

PR #826, opened 2026-05-20
Foreign-key dependent fix Project assets, causing issue in reseed task and endpoint (integration)

abcampo-iry:
## Status

- Related to https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1419

## Points for consideration:

- Security
- Performance

## What's changed?

Fix /test/reseed failures caused by Scratch project assets blocking seeded project cleanup.
The test seed destroy task removes seeded lesson projects with Project.destroy_all, but project-scoped ScratchAsset records still referenced those projects.

Added a test and found is red with same error as sentry.

I only wonder here maybe if we want to make dependent :destroy or dependent nullify. (I guess this doesn't happen manually but it's worht to think if we would like to solve it in other way)


zetter-rpf (approved):
Nice one

zetter-rpf:
> I only wonder here maybe if we want to make dependent :destroy or dependent nullify. (I guess this doesn't happen manually but it's worht to think if we would like to solve it in other way)

I think it's hard to decide because we don't have any use cases for deleting scratch projects in production and we're only doing it in test. Think this is a great approach for now.

PR #849, opened 2026-06-04
Scratch projects can be copied fix

abcampo-iry:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1465

## What's changed?

The bug was that lesson copy already duplicated the lesson/project shell, but Scratch projects also need two Scratch-specific things:

 1. scratch_component.content
 2. project-owned scratch_assets

Without those, the copied project exists, but Scratch cannot load the project/asset data correctly.

https://github.com/user-attachments/assets/bb0aed59-69f7-4ba0-abfb-1d3ec06ad495


zetter-rpf (approved):
Nice one, very clear

PR #851, opened 2026-06-05
Expand Salesforce sync to SchoolClass, ClassTeacher, and Lesson

fspeirs:
Related to https://github.com/RaspberryPiFoundation/experience-cs/issues/2142

## Summary

Extends the Heroku Connect Salesforce sync pattern from #677 to three more Salesforce objects:

| editor-api model | Salesforce object | Fields synced |
| --- | --- | --- |
| `SchoolClass` | `Classroom__c` | identifiers, name, member count, parent `Editor__c` lookup |
| `ClassTeacher` | `Contact_Classroom_Affiliation__c` | identifiers, parent `Classroom__c` and `Contact` lookups |
| `Lesson` (classroom-bound) | `Lesson__c` | identifiers, name, teacher-project metadata, assigned/completed counts, parent `Classroom__c` lookup |

It also generalises the parent-sync race guard that #850 introduced for `RoleSyncJob` so every sync job using `__r__<external_id>` lookups gets it for free, and documents the operational shape of the sync layer in `AGENTS.md`.

## Behaviour

**Triggers.** `after_commit` on each source model enqueues a per-record GoodJob (`Salesforce::{SchoolClass,Lesson,ClassTeacher}SyncJob`), gated on `FeatureFlags.salesforce_sync?`. Per-record concurrency on the base job collapses duplicate enqueues from same-transaction commits (e.g. new class + initial teacher).

**Lesson completion is two flows.** `Lesson__c.numberofcompletedprojects__c` is the sum of:
- `Lesson#submitted_projects_count` — Statesman-driven, used by the Code Editor (Python/HTML) flow when a `SchoolProject` transitions to `:submitted`. Unchanged from main.
- `Lesson#finished_projects_count` — live read of `school_projects.where(finished: true).count`, used by Experience CS for Scratch projects (which flip `finished` directly and never enter the state machine).

The two flows are mutually exclusive in practice, so summing is safe. `SchoolProject` gets a new `after_commit` on `saved_change_to_finished?` to enqueue `LessonSyncJob` when Experience CS flips the flag.

**Assignment count is event-driven.** `Lesson__c.numberofassignedprojects__c` is `lesson.remixes.count`, which only changes when a remix `Project` is created. The enqueue lives on `Project after_commit :create` (`remixed_from_id.present?`), not on roster changes — so adding a student to a class no longer scans every lesson in the class.

**Parent-sync race guard.** `ensure_parent_synced!(model, external_id_field, external_id, label)` on `Salesforce::SalesforceSyncJob` checks the parent has a non-nil `sfid` in its Heroku Connect mirror and raises `SalesforceRecordNotFound` if not. The base job retries with polynomial backoff, so jobs self-heal once the parent lands instead of leaving the child row permanently `FAILED` in Heroku Connect. This is a generalisation of the mechanism introduced in #850.

## Backfill

After deploying with `SALESFORCE_ENABLED=true`, run the backfill rake tasks **in this order** so each child's parent already has an `sfid` in its mirror by the time the child enqueues:

```bash
rails salesforce_sync:school
rails salesforce_sync:role
rails salesforce_sync:contact
rails salesforce_sync:school_class
rails salesforce_sync:class_teacher
rails salesforce_sync:lesson
```

Each task is `Model.find_each { perform_later }` — non-blocking; the worker drains the `salesforce_sync` queue at concurrency 10. Watch progress at `/admin/good_job`.

Strict ordering isn't load-bearing: `ensure_parent_synced!` defers any child whose parent hasn't landed via `retry_on SalesforceRecordNotFound, wait: :polynomially_longer, attempts: 10`, so out-of-order runs self-heal. Running parents first just cuts retry churn.

Re-running is idempotent — each job uses `find_or_initialize_by` on the external-ID key, so repeats update in place rather than duplicating rows.

## Out of scope

- **Destroy semantics for mirror rows.** This PR doesn't delete or soft-delete mirror rows when the source record is destroyed (existing behaviour). Worth deciding before GA.
- **Observability on race-guard exhaustion.** No Sentry alert when `retry_on SalesforceRecordNotFound` exhausts its attempts.
- **`salesforce_sync:all` convenience task** to run the canonical backfill order in one go.

## Test plan

- [ ] `docker compose run --rm api rspec spec/jobs/salesforce spec/models/{lesson,school_class,school_project_salesforce_sync,class_teacher_salesforce_sync,class_student_salesforce_sync,project_salesforce_sync}_spec.rb`
- [ ] `docker compose run --rm api bundle exec rubocop`
- [ ] After deploy with `SALESFORCE_ENABLED=true`, run the backfill tasks per the Backfill section above.
- [ ] In `salesforce_connect`, confirm every parent row has a non-nil `sfid` before child writes and that any transient `_hc_err` rows self-heal via the retry.

Made with [Cursor](https://cursor.com)


app/models/lesson.rb
```diff
@@ -41,8 +44,23 @@ def recalculate_submitted_projects_count!
...
   end
 
+  # Counts Experience CS Scratch completions for this lesson. Experience CS marks a
+  # project complete by flipping school_projects.finished via Concept::SchoolProject::SetFinished,
+  # bypassing the Statesman state machine entirely — so submitted_projects_count never
+  # sees these. Read live (no cached column) because the only consumer is LessonSyncJob,
+  # which is per-lesson concurrency-limited; if a non-sync reader ever needs this, add a
+  # cached column then. Summed with submitted_projects_count in LessonSyncJob to derive
+  # Lesson__c.numberofcompletedprojects__c.
+  def finished_projects_count
```
zetter-rpf: I think this comment is a bit of an information overload and not sure how helpful it is.

- why would a non-sync reader need to use a cached column? I would expect this to be doing a single database query, what's the problem with that?
- I think it's unusual to list callers within documentation - if this changes then the comments not going to be kept in date.

Instead of this comment, I wondered if there's a way to make it clearer that this is for experience CS projects - for example just calling it `finished_experience_cs_scratch_project_count` would make that clear and maybe make the comment uneeded.

app/jobs/salesforce/lesson_sync_job.rb
```diff
@@ -0,0 +1,55 @@
...
+      sf_lesson.save!
+    end
+
+    private
+
+    def sf_lesson_attributes(lesson:)
+      mapped_attributes(lesson:).merge(
+        teacherprojecttitle__c: lesson.project&.name,
+        teacherprojecttype__c: lesson.project&.project_type,
+        numberofassignedprojects__c: lesson.remixes.count,
```
zetter-rpf: Is counting remixes a good way to tell us assigned projects?

Remixes count might be a good approximation for _started_ projects, but it wouldn't tell you if a teacher has assigned a project. As far as I know we don't model project assignment, we just know if a project has been made available to a student in a class.

This might be a naming thing - maybe `numberofassignedprojects__c` should be called 'number of started/saved projects' or similar

app/models/class_teacher.rb
```diff
@@ -18,12 +18,27 @@ class ClassTeacher < ApplicationRecord
...
 
   private
 
+  def enqueue_school_class_sync
+    Salesforce::SchoolClassSyncJob.perform_later(school_class_id: school_class_id)
+  end
+
+  def enqueue_class_teacher_sync
+    Salesforce::ClassTeacherSyncJob.perform_later(class_teacher_id: id)
+  end
```
zetter-rpf, line 39: Does this work with owners and does it need to? I think owners can manage all classes but I don't know if they have a class teacher record or not.

zetter-rpf:
I'm finding some of the comments in app/models files hard to understand, I think it's because they are terminology that's not in our codebase already (like 'roster', or 'parent classroom mirror'). I think it's also because they leak some of the details out of the salesforce implementation into the models.

Do you think these comments are important? I wasn't sure if you've taken the time to write them or they were autogenerated. Some of these might be better documented in the PR or commit history.

spec/support/negated_matchers.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# Negated forms of common matchers, for use inside compound expectations like
+# `.to have_enqueued_job(A).and not_have_enqueued_job(B)`.
+RSpec::Matchers.define_negated_matcher :not_have_enqueued_job, :have_enqueued_job
```
zetter-rpf, line 5: Nice!

app/jobs/salesforce/lesson_sync_job.rb
fspeirs: Yeah, I think it's a terminology issue, maybe. The assumption is that if the project is assigned to the class, it is assigned to all the students in the class.

app/models/lesson.rb
fspeirs: I did consider putting some reference to `experience_cs` in the method name, but decided against it on the grounds that `finished` is a first-class schema field in `editor-api`, which leads to a bit of necessary explanation in the comment (I agree that it can be simplified and I will).

Let me know what you think - I'm happy to use your function name if you think it will be clearer going forward.

app/jobs/salesforce/lesson_sync_job.rb
zetter-rpf: So maybe it could be removed if the existing class member count is sufficient?

app/models/class_teacher.rb
fspeirs: It'll work for anyone who has the teacher role for the school that owns the classroom. As I understand the requirements here, it's about people who are actively administering the classroom, rather than "people who have permissions to view the classroom".

fspeirs:
On the question of leaking Salesforce into the models, I somewhat agree. One of my concerns here, that I rather ran out of time to consider doing, is that this work layers a network of event-driven sync jobs, where the dependencies between them might be rather non-obvious to future developers.

app/state_machines/school_project_state_machine.rb
```diff
@@ -17,4 +17,6 @@ class SchoolProjectStateMachine
 
   after_transition(to: :submitted, &:recalculate_lesson_submitted_projects_count!)
   after_transition(from: :submitted, &:recalculate_lesson_submitted_projects_count!)
+  after_transition(to: %i[submitted complete], &:enqueue_salesforce_lesson_sync)
+  after_transition(from: :complete, &:enqueue_salesforce_lesson_sync)
```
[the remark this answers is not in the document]
fspeirs: Fixed.

app/models/class_teacher.rb
zetter-rpf: I checked and it looks like we do create class_teacher records for school owners (I'm not sure where in the code we do this, but it works on staging) so this is fine.

app/jobs/salesforce/lesson_sync_job.rb
fspeirs: Apologies, I mis-remembered how that part worked. What counts as an "assigned" project is when the student has first saved their remix of a teacher's project.

zetter-rpf: I saw you added a comment to the code, but I'm unsure about the reasoning behind this - what do the salesforce team want to track - project assignments or project starts?  

If it's project assignments, shouldn't use the class member count (or remove this since we already have that?). if it's project starts, could this be renamed? Or is this terminology already used for reporting Experience CS metrics?

zetter-rpf (approved):

zetter-rpf:
Great, thanks for the improvements

app/jobs/salesforce/lesson_sync_job.rb
fspeirs: We discussed this on Slack but, for the GitHub record, we decided that the actual metric is "projects visible to students in the classroom" so in commit aca52ff67ac6108e161959abcb10aa35fcfeb830 we changed it to this calculation.

PR #853, opened 2026-06-08
Bulk lesson creation

PetarSimonovic:
## Status

- Closes [#1472](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1472#issue-4611889666)

- Related to [Experience CS #1761 ](https://github.com/RaspberryPiFoundation/experience-cs/issues/1761)

## What's changed?

- Add Lessons::Batch controller 
- Add a bulk create jbuilder that renders an array, echoing an `origin_identifier` when supplied so the caller can match responses to submitted projects 
- Add a LessonsCreation concern to consolidate logic and params that are share between batch and single lesson creation
- Update specs


app/views/api/lessons/create_bulk.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+json.array!(@results) do |result|
+  if result.success?
+    lesson, user = result[:lesson].with_user
+    json.partial! 'lesson', lesson: lesson, user: user
```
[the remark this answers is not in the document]
PetarSimonovic: Implemented

app/controllers/api/lessons_controller.rb
```diff
@@ -30,13 +47,19 @@ def show
...
-      result = Lesson::Create.call(lesson_params: create_params)
-
-      if result.success?
-        @lesson_with_user = result[:lesson].with_user
-        render :show, formats: [:json], status: :created
+      if params[:lesson_projects].present?
+        @results = Lesson::CreateBulk.call(
+          lessons_params: params[:lesson_projects].map { |entry| bulk_create_params(entry) }
+        )
+        render :create_bulk, formats: [:json], status: :created
```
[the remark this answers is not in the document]
PetarSimonovic: I've implemented this change using the existing `verify_can_create_scratch_projects` hook

app/controllers/api/lessons_controller.rb
```diff
@@ -4,6 +4,23 @@ module Api
   class LessonsController < ApiController
```
zetter-rpf, line 4: This controller is getting longer now and I think having different code paths for the create action and many of the helpers does make it harder to follow and easier to introduce bugs (although I like having the attributes shared between them as it means that they are updated together).

Do you think it's worth improving this as part of this change? I wondered if it would be better or worse to have a different bulk controller, or if other behaviour could be moved outside of the controller.

spec/features/lesson/creating_a_lesson_spec.rb
```diff
@@ -118,6 +118,170 @@
...
+        school_class.update!(teachers: [ClassTeacher.new({ teacher_id: teacher.id })])
+      end
+
+      it 'responds 201 Created' do
+        post('/api/lessons', headers:, params: { lesson_projects: lesson_project_params })
+
+        expect(response).to have_http_status(:created)
+      end
+
+      it 'responds 422 Unprocessable if school_class_id does not correspond to school_id' do
```
zetter-rpf: What happens if the school_class matches the school_id but neither are associated with the current user?



I haven't confirmed if this works or not in the existing create action - Cancancan should authorise the resource in create actions by creating a new record with the params  - `create_params` or `lesson_params` here but this won't work with `bulk_create_params`. ([see cancancan docs](https://redirect.github.com/CanCanCommunity/cancancan/blob/develop/docs/controller_helpers.md#strong-parameters)).

app/controllers/api/lessons_controller.rb
PetarSimonovic: I've separated bulk creation into its own controller. 

I've also created a LessonCreation concern for params and logic that are shared between single and batch lesson creation.

Let me know if LessonCreation makes it harder to follow the behaviour and I can move those methods back into the controllers.

spec/features/lesson/creating_a_lesson_spec.rb
PetarSimonovic: I've added a `before_action` to loop through and `authorize!` each lesson. I've updated the specs to check that the controller responds with 403 if the user does not belong to the school.

zetter-rpf (approved):
I found the batch controller much easier to understand, thanks for splitting that out.

PR #855, opened 2026-06-08
Importing projects site projects from .sb3 files

### LINKED ISSUE #845: Handle .sb3 files in UploadJob
- Pass a single sb3_file: { filename:, io: } into the importer (convention: main.sb3).

See [ADR](https://digital-docs.rpf-internal.org/docs/teams/code-club/adr/2026-05-18-scratch-integration-projects-site/) for more context.

### LINKED ISSUE #846: Extend `ProjectImporter` for Scratch
Extend `ProjectImporter` or create a new `ScratchStarterImporter`:

- gate this on projects with type `code_editor_scratch` only.
- skip `delete_components` and `create_components`
- Upsert into `scratch_components`, set content from parsed JSON.
- Upsert into `scratch_assets`

### LINKED ISSUE #847: Create Sb3Parser
- Unzip the archive into project.json plus assets
- Parse project.json into a Hash for `scratch_component.content`.
- Collect asset entries whose names match md5ext values referenced in the JSON.

Probably no need to enforce a size limit at this stage.

rammodhvadia:
## What's changed?

- Added an sb3 parser to extract scratch component json and asset files from `.sb3` file
- Added new functions to import assets from sb3 files in ProjectImporter
- Updated ProjectImporter to import scratch component and asset for `code_editor_scratch` projects
- Updated UploadJob to process `.sb3` files
- Updated FileSystemProject to accept `.sb3` files


lib/scratch_asset_importer.rb
```diff
@@ -55,6 +67,17 @@ def create_scratch_asset
...
+    return if ScratchAsset.global_assets.exists?(filename: asset_name)
+
+    sleep(ASSET_FETCHING_DELAY)
+    ScratchAsset.create!(filename: asset_name, project_id: nil, uploaded_user_id: nil)
+                .file
+                .attach(io: StringIO.new(content), filename: asset_name)
+  rescue StandardError => e
+    Rails.logger.error("Failed to import SB3 asset #{asset_name}: #{e.message}")
+  end
+
```
zetter-rpf: I'm wondering if this is the best place for these methods-  this file was specific to importing assets from the Scratch asset library and it looks like the new code doesn't re-use the existing methods. Having multiple distinct call paths through a file makes it harder to understand and test.

There's also some patterns in this method that make sense for the original file but I don't think are right here:
- Using `sleep` - this was added before as we were fetching from a 3rd party API which we're not here (we are saving to s3 but that can handle the volume)
- Rescuing the error and logging - this was done before because the code is run from a rake task so an engineer can see and respond to errors. For this I'm I wonder if it would be better **not** to rescue so errors are collected in sentry.

It might be clearer to create a new file that's more specific (`scratch_sb3_asset_importer` ?) and have the functionality in there. You could even name this file to the more specific `scratch_asset_library_importer`.

spec/support/sb3_archive_helper.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Sb3ArchiveHelper
```
zetter-rpf, line 4: I like this helper, very useful for these tests!

jamdelion (approved):
Looks good to me, but may be worth waiting for a Ruby expert's eye just in case 👁️

lib/tasks/project_components/scratch-integration-test-starter/project_config.yml
```diff
@@ -0,0 +1,9 @@
+NAME: "scratch integration test"
+IDENTIFIER: "editor-scratch-testing-starter"
+TYPE: "code_editor_scratch"
+COMPONENTS:
+  - name: "main"
+    extension: "sb3"
+    location: "main.sb3"
+    index: 0
+    default: true
```
zetter-rpf: Will there only ever be one component for Scratch projects?

I just wondered if there is use in having this here if it will be the same for every Scratch project. Allowing multiple components might also lead to odd behaviour.

lib/project_importer.rb
```diff
@@ -39,16 +41,42 @@ def setup_project
...
     components.each do |component|
       project_component = Component.new(**component)
       project.components << project_component
     end
   end
 
+  def create_scratch_component
+    return unless project.project_type == 'code_editor_scratch'
+
+    components.each do |component|
```
zetter-rpf: Similar to my comment about the example project_config - this `each` here makes it look like we expect scratch files to have multiple components - do we?

If a Scratch project has multiple components what behaviour do we want - to try to import them all? To raise an error?

lib/project_importer.rb
```diff
@@ -39,16 +41,42 @@ def setup_project
...
+      next unless component[:extension] == 'sb3'
+
+      parsed_content = Sb3Parser.new(component: component).parse.fetch(:scratch_component).fetch(:content)
+      project.scratch_component = ScratchComponent.new(content: parsed_content) if parsed_content
```
zetter-rpf: When will `parsed_content` be falsy? And would you want it to raise in this case or continue?

zetter-rpf (commented):
Great work, for me the main thing that would be good to improve is the changes to `lib/scratch_asset_importer.rb‎` file the and error handling in it. Left a few other minor comments/questions.

lib/scratch_asset_importer.rb
rammodhvadia: Agreed about the splitting it out, have updated

lib/project_importer.rb
rammodhvadia: Should only ever be one component - the sb3 file. Have removed the `each` and only handling 1 component now to make that clearer

app/jobs/upload_job.rb
```diff
@@ -131,6 +131,11 @@ def categorize_files(files, project_dir, locale, repository, owner)
     }
 
     files.each do |file|
+      if file.extension == '.sb3'
+        categories[:components] << scratch_file_component(file, project_dir, locale, repository, owner)
+        next
+      end
```
patch0, line 137: I'd be tempted to add the `sb3` as an extra mime type to Marcel (the library used to determine mime types) and then put this code in the case statement.

Create an initializer like this to configure Marcel
```ruby
Marcel::MimeType.extend "application/x.scratch.sb3", extensions: %w[sb3]
```

and then add

```ruby
when 'application/x.scratch.sb3'
  categories[:components] << scratch_file_component(file, project_dir, locale, repository, owner)
```

Keeps things more as-is, IMO, but it (might) have the side effect of making any scratch file downloads come with the correct mime type too.

app/jobs/upload_job.rb
```diff
@@ -158,11 +163,20 @@ def component(file)
...
 
+  def scratch_file_component(file, project_dir, locale, repository, owner)
+    name = file.name.chomp(file.extension)
+    extension = file.extension[1..]
+    { name:, extension:, io: URI.parse(file_url(file, project_dir, locale, repository, owner)).open }
+  end
+
   def media(file, project_dir, locale, repository, owner)
     filename = file.name
+    { filename:, io: URI.parse(file_url(file, project_dir, locale, repository, owner)).open }
```
patch0: (nitpick) I'd not have the extra assignment here
```suggestion
    { filename: file.name, io: URI.parse(file_url(file, project_dir, locale, repository, owner)).open }
```

app/jobs/upload_job.rb
```diff
@@ -158,11 +163,20 @@ def component(file)
...
+  end
+
+  def file_url(file, project_dir, locale, repository, owner)
     directory = project_dir.name
-    url = "https://github.com/#{owner}/#{repository}/raw/#{ENV.fetch('GITHUB_WEBHOOK_REF')}/#{locale}/code/#{directory}/#{filename}"
-    { filename:, io: URI.parse(url).open }
+    "https://github.com/#{owner}/#{repository}/raw/#{ENV.fetch('GITHUB_WEBHOOK_REF')}/#{locale}/code/#{directory}/#{file.name}"
```
patch0: Or here, just use `project_dir.name` in the URL string.

app/jobs/upload_job.rb
patch0: Actually, looking at `app/models/filesystem_project.rb` the extension is used before the mime type.  Lets leave this as-is.

app/models/filesystem_project.rb
```diff
@@ -53,9 +54,18 @@ def self.categorize_files(files, dir)
     categories
   end
 
+  def self.configured_scratch_files(files, proj_config)
+    configured_locations = Array(proj_config['COMPONENTS']).pluck('location')
+    return files if configured_locations.empty?
+
+    files.reject { |file| File.extname(file) == '.sb3' && configured_locations.exclude?(file.basename.to_s) }
```
patch0: What does this line do?  I notice it isn't covered by tests.

rammodhvadia: Was used to check the project config and only include sb3 files listed there, but as we are only checking for 1 sb3 file, it's probably safe to remove - will update

lib/sb3_parser.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  private
+
+  def open_zip(&)
+    return Zip::File.open(file_path, &) if file_path
+
+    io.rewind if io.respond_to?(:rewind)
+    result = nil
+    Zip::File.open_buffer(io.read) { |zip_file| result = yield zip_file }
+    result
```
patch0: Is the `rewind` needed?  Also you don't need to assign the result, but you can call `return` inside the block.

```suggestion
    Zip::File.open_buffer(io.read) do
      return yield it
    end
```

mwtrew (approved):
Changes since Chris' review look fine to me.

PR #866, opened 2026-06-12
Ensure safeguarding flags before Profile student calls

abcampo-iry:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1475

## What's changed?

This change ensures safeguarding flags are created before editor-api calls Profile’s protected school-student endpoints.

Instead of relying on individual controllers to create the flags, flag creation now happens at the `SchoolStudent` / Profile API boundary. This covers controller and non-controller callers, including class member listing, project student-name lookups, SSO student imports, batch student creation jobs, and student removal.

The duplicated safeguarding flag logic has been removed from `SchoolStudentsController` and `SchoolMembersController`, and a new `SafeguardingFlagService` centralizes owner/teacher flag creation.

Member listing now also propagates `SchoolStudent::List` failures instead of treating failed student lookups as empty lists. This prevents school/class member endpoints from returning successful but incomplete responses when safeguarding flag creation or Profile API calls fail.

### DIFF
--- a/app/services/student_removal_service.rb
+++ b/app/services/student_removal_service.rb
@@ -12,6 +12,8 @@ class StudentRemovalService
   def remove_students
     results = []
 
+    ensure_safeguarding_flag if remove_from_profile?
+
     @students.each do |user_id|
       student_roles = Role.student.where(user_id:, school_id: @school.id)
       if student_roles.empty?
@@ -21,8 +23,6 @@ class StudentRemovalService
 
       result = { user_id: }
       begin
-        ensure_safeguarding_flag if remove_from_profile?
-
         ActiveRecord::Base.transaction do
           # Delete all projects for this user
           projects = Project.where(user_id: user_id)
@@ -53,10 +53,7 @@ class StudentRemovalService
   end
 
   def ensure_safeguarding_flag
-    return if @safeguarding_flag_created
-
     SafeguardingFlagService.create_for_token(token: @token, school: @school)
-    @safeguarding_flag_created = true
   end
 
   def remove_from_profile?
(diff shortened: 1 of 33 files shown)


app/services/student_removal_service.rb
```diff
@@ -21,6 +21,8 @@ def remove_students
 
       result = { user_id: }
       begin
+        ensure_safeguarding_flag if remove_from_profile?
```
mwtrew: I suppose it depends how strict we want to be with associating a safeguarding flag with a particular operation, but wouldn't it be simpler to attempt to set the safeguarding flag early, outside the loop? 

My understanding is that the safeguarding flag indicates a user has changed the school in some way - so if we could think of this batch delete as a single "change" we could simplify the code.

mwtrew (approved):
Looks good overall, and I think the RequestStore caching to prevent repeated Hydra calls is nice. A couple of questions:

Did you consider the suggestion in the issue to bring safeguarding flag creation forward to role-creation-time?

>Ideally the flag in profile should be created as soon as the role is created in the editor-api. As we do have users that have roles in the editor but not in profile, these should either be backfilled in after that fix is in place, or the workaround to attempt to create the flags needs to be extended to cover all the api endpoints.

(I see you've gone for the latter)

app/services/student_removal_service.rb
abcampo-iry: Yes, agreed. That is simpler.
The flag is for the current user making a school change, not for each student being deleted, so the batch delete can be treated as one change.

I’ll move the safeguarding flag creation before the loop and remove the memo guard.

PR #899, opened 2026-06-26
Add extra required data to the track_event call when creating a school

jamiebenstead:
## Status

- [Closes](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1485)
- [Related to](https://github.com/RaspberryPiFoundation/editor-standalone/pull/961)

### Description of what's been done
- Add `first_landing_page` and `marketing_preference` data when saving the School - Created event


app/controllers/api/schools_controller.rb
```diff
@@ -20,7 +20,12 @@ def create
 
       if result.success?
         @school = result[:school]
-        track_event('School - Created', school_id: @school.id)
+        track_event(
+          'School - Created',
+          school_id: @school.id,
+          first_landing_page: params[:first_landing_page],
+          marketing_parameters: marketing_parameters
```
zetter-rpf: I'm wondering if it would be easier to query in the data warehouse if all of the event properties were at the same level. It might just be harder to document and query if some are nested inside other attributes. This means we could query on `events.properties['utm_source']` rather than `events.properties['marketing_params']['utm_source']`.

You should be able to do this by spatting the marketing parameters: 

`track_event('School - Created', **marketing_parameters, @school.id:, first_landing_page: params[:first_landing_page])`

zetter-rpf (approved):
Great. added a suggestion - let me know if you think it's worth doing.

zetter-rpf (commented):
The track_event call looks good
I saw the comments from cursor and copilot - do you think they could be problems?

zetter-rpf (approved):
Nice one

PR #900, opened 2026-06-29
Restrict teacher ability so that they can only access their own class projects

cocomarine:
## Status

- Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/526

## What's changed?

- Fixed a bug where teachers could access projects in school classes they don't belong to by adding class membership check to the project rule in school teacher abilities.


zetter-rpf (approved):
Nice one

PR #901, opened 2026-06-29
Project importer fixes

rammodhvadia:
## What's changed?

- Fixes an issue with the `skip_job` variable. Once `@skip_job` was set to true by a project with `build: false` in its `project_config.yml`, it stayed true for the rest of the run, meaning every following project import for every locale would be skipped.
- Adds a test to check the old behaviour does not happen


patch0 (approved):
🕵🏻 🎉

zetter-rpf (approved):
Nice one

PR #938, opened 2026-07-24
Stop reporting expected student validation errors to Sentry

abcampo-iry:
Related to Sentry issue https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1583

## Points for consideration:

- No security impact.
- No performance impact.

## What's changed?

- When someone tries to create a student with:
  - Username already taken
  - Password matches the username
  - Password too short
  - Missing or invalid student details

Profile API returns an isComplex validation error.
- This is an expected error that is shown to the user, but we were also sending it to Sentry.
- This change handles student validation errors separately, so they are no longer sent to Sentry (since i don't think they add value and cause noise)..

No additional steps are required. There are no migrations, configuration changes, or data backfills.

### DIFF
--- a/lib/concepts/school_student/create.rb
+++ b/lib/concepts/school_student/create.rb
@@ -7,6 +7,9 @@ module SchoolStudent
         response = OperationResponse.new
         response[:student_id] = create_student(school, school_student_params, token)
         response
+      rescue ProfileApiClient::Student422Error => e
+        response[:error] = e.to_s
+        response
       rescue StandardError => e
         Sentry.capture_exception(e)
         response[:error] = e.to_s
--- a/lib/concepts/school_student/create.rb
+++ b/lib/concepts/school_student/create.rb
@@ -4,16 +4,13 @@ module SchoolStudent
   class Create
     class << self
       def call(school:, school_student_params:, token:)
-        response = OperationResponse.new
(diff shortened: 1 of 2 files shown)


lib/concepts/school_student/create.rb
```diff
@@ -4,13 +4,13 @@ module SchoolStudent
...
-        response
+        student_id = create_student(school, school_student_params, token)
+        OperationResponse[student_id:]
+      rescue ProfileApiClient::Student422Error => e
+        OperationResponse[error: e.to_s]
       rescue StandardError => e
         Sentry.capture_exception(e)
-        response[:error] = e.to_s
-        response
+        OperationResponse[error: e.to_s]
```
zetter-rpf, line 13: Doesn't need to be changed as part of this, but I'm unsure if there is value in having this here since there's already something in the api controller that converts errors to an API response. I think this would only be useful if there there is a need in this controller that this is used to render errors in a different format to what the base api controller does.

zetter-rpf (approved):
Nice one

lib/concepts/school_student/create.rb
abcampo-iry: That makes sense. We can create a separate ticket to discuss the exact cleanup later.

PR #952, opened 2026-08-06
Address potential weaknesses source of join-code generation

abcampo-iry:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1582

## What's changed?

My guess for the Sentry error is that, JoinCodeGenerator created one Random.new instance when the application loaded. 

Maybe Puma preloads the application and then forks its workers, so each worker inherited the same random-generator state. Those workers could therefore generate the same sequence of join codes, or there is a hash collision because of not being random although this is less likely.

The PR replaces that shared Random instance with SecureRandom, which obtains fork-safe randomness from the operating system. Each worker therefore generates an independent sequence.

### DIFF
--- a/lib/join_code_generator.rb
+++ b/lib/join_code_generator.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'securerandom'
+
 class JoinCodeGenerator
   # Omit K, X, Z — commonly confused or offensive in short codes.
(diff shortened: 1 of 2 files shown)


zetter-rpf (approved):

zetter-rpf:
I like your theory! lets ship and see if it is fixed

PR #953, opened 2026-08-07
Support instruction steps

zetter-rpf:
## Status

- Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1681

## What's changed?

+ Added support for instructions to contain multiple steps 
+ Simplified how we were preventing students from updating instructions

The new instructions will be in a format such as:

```
[
  {markdown_content: 'step 1'},
  {markdown_content: 'step 2'},
]
```

This has been added to a new `instruction_steps` column that will eventually replace instructions.

This change is backwards compatible with current versions of editor since sole strings are allowed in the JSONB column. It won't be at the point that we start storing instruction steps instead of strings in that column.

*Created with help from Claude Code*

### DIFF
--- a/db/migrate/20260807120000_add_instruction_steps_to_projects.rb
+++ b/db/migrate/20260807120000_add_instruction_steps_to_projects.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddInstructionStepsToProjects < ActiveRecord::Migration[8.1]
+  def change
+    add_column :projects, :instruction_steps, :jsonb
+  end
+end
--- a/lib/concepts/project/operations/update.rb
+++ b/lib/concepts/project/operations/update.rb
@@ -45,8 +45,15 @@ class Project
       def student_project_instructions_updated?(response, update_hash, current_user)
         is_school_project = response[:project].school.present?
         user_is_student = current_user.student?
-        instructions_updated = response[:project].instructions != update_hash[:instructions]
+        instructions_updated = normalize_instructions(response[:project].instructions) != normalize_instructions(update_hash[:instructions])
         is_school_project && user_is_student && instructions_updated
+      end
+
+      # update_hash[:instructions] may be plain Ruby data or an ActionController::Parameters
+      # array/hash (for the instruction steps format); round-trip through JSON so both sides
+      # of the comparison are in the same plain, comparable shape.
+      def normalize_instructions(value)
+        ActiveSupport::JSON.decode(ActiveSupport::JSON.encode(value))
       end
 
       def validate_update(response, update_hash, current_user)
--- a/lib/concepts/project/operations/update.rb
+++ b/lib/concepts/project/operations/update.rb
@@ -3,11 +3,11 @@
 class Project
   class Update
     class << self
-      def call(project:, update_hash:, current_user:)
+      def call(project:, update_hash:)
         response = setup_response(project)
 
         setup_deletions(response, update_hash)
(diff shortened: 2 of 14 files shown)


db/migrate/20260807120000_change_projects_instructions_to_jsonb.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ChangeProjectsInstructionsToJsonb < ActiveRecord::Migration[8.1]
+  def up
+    # to_jsonb(text) wraps the existing markdown strings as JSON string scalars,
+    # so plain-string instructions keep reading back exactly as they were.
+    change_column :projects, :instructions, :jsonb, using: 'to_jsonb(instructions)'
+  end
```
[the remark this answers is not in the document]
zetter-rpf: Yeah I was worried about this, I think it's fine if it it takes 30s or so, any longer could be disruptive. Unsure if I should break it up to deploy in steps or just deploy at a less busy time.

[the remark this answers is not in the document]
zetter-rpf: There's now a new column

db/migrate/20260807120000_add_instruction_steps_to_projects.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddInstructionStepsToProjects < ActiveRecord::Migration[8.1]
+  def change
+    add_column :projects, :instruction_steps, :jsonb
```
jamiebenstead, line 5: What made you decide to add the extra column rather than modify and use the current instructions column?

zetter-rpf: Good question, there's some details in the commit message:

> Changing the existing instructions column type in place was tried
first, but that is risky against a live database: a server process
that queried the table before the migration ran keeps the old
column type cached and keeps sending data the database no longer
accepts, causing errors until it is restarted. Converting existing
data to the new type can also take a while on a large table and
holds a lock for that duration, causing errors on any concurrent
read or write.

lib/concepts/project/operations/update.rb
```diff
@@ -3,11 +3,11 @@
 class Project
   class Update
     class << self
-      def call(project:, update_hash:, current_user:)
+      def call(project:, update_hash:)
```
jamiebenstead, line 6: I'm curious, why are we removing the `current_user` everywhere here?

db/migrate/20260807120000_add_instruction_steps_to_projects.rb
zetter-rpf: This isn't true for all type changes, but was for this one when I tested it because Rails treats jsonb differently.

jamiebenstead (approved):
LGTM. Left a few comments on things I am curious about

lib/concepts/project/operations/update.rb
zetter-rpf: There's detail in the commit message:

> Previously Project::Update compared old and new instructions to
detect and reject a student trying to change them, needing a
current_user argument and a JSON-normalising comparison just to
tell Parameters and plain Ruby values apart.

I think it's simpler to do this in the controller and let Rails manage unexpected parameters rather than special casing this. Having it in the update required more complexity to see if the instructions had changed which was added in the first commit.

PR #959, opened 2026-08-11
Allow Experience CS to sync public Scratch projects

abcampo-iry:
## Status

- Partially closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1657
- Related to https://github.com/RaspberryPiFoundation/experience-cs/pull/2330

  ## Points for consideration:

- Security: Only users with the `experience-cs-admin` role can change the project type and Scratch data for public projects.
- Security: Projects are not yet marked as coming from Experience CS. Until origin tracking is added, only known Experience CS project identifiers should be enabled for synchronization.

## What's changed?

- Allow public, localized `code_editor_scratch` projects to contain instructions. This is a narrow exception to [Restrict instructions to school projects (#479)](https://github.com/RaspberryPiFoundation/editor-api/pull/479) for public Experience CS templates. This may be worth discussing.
- Allow Experience CS admins to change a public project to `code_editor_scratch`.
- Allow Experience CS admins to create or update Scratch data through the project update endpoint.

Some concerns of mine were to understand the cases:

- Existing ExCS stub with no ScratchComponent -> Create ScratchComponent -> Store the ExCS code -> Change type from scratch to code_editor_scratch

- Existing ExCS project with a ScratchComponent -> Replace its content with the latest ExCS code  -> This is intentional because ExCS is the source of truth

- User-owned project  -> Cannot be updated by an experience-cs-admin  -> Authorization prevents the overwrite

- Unrelated public project with the same identifier  -> Could currently be overwritten  -> This is the known provenance/origin risk mentioned in the ticket

### DIFF
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -104,18 +104,25 @@ module Api
         :user_id,
         :identifier,
         :name,
-        :project_type,
         :locale,
         {
           components: %i[id name extension content index default]
         },
-        scratch_component: {},
         parent: {},
         image_list: []
       ]
-      return attributes if current_user&.student?
+      attributes.push(:instructions, { instructions: [:markdown_content] }) if can_set_instructions?
+      attributes.push(:project_type, { scratch_component: { content: {} } }) if can_set_project_type_and_scratch_data?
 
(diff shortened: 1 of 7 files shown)


zetter-rpf (approved):
Nice one

PR #962, opened 2026-08-12
Surface the underlying error when school onboarding fails

jamiebenstead:
Previously SchoolOnboardingService caught every exception and returned false, so School::Create raised a bare 'School onboarding failed' RuntimeError in its place. Sentry grouped every possible cause under that one message - a Profile API timeout, a 5xx, a role that could not be created - with a backtrace pointing at the raise rather than at whatever actually broke. Non-401 failures were also reported twice, once from the service and again as the RuntimeError.

This change lets the exception propagate instead. The existing rescue in School::Create receives the real error, so Sentry records its class, message and backtrace once, and the generic RuntimeError is gone entirely.

The ProfileApiClient::UnauthorizedError rescue moves up to School::Create so that Profile 401s, which happen when a user is not yet verified, still log a warning without being sent to Sentry. Those registrations now produce no Sentry event at all, so the 'user is unauthorized' log line is the only trace of them.

The operation response keeps its shape, so the API still returns 422, but the message is the underlying error rather than 'School onboarding failed'.

Co-Authored-By: Claude Opus 5

## Status

- closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1600


lib/concepts/school/operations/create.rb
```diff
@@ -10,21 +10,28 @@ def call(school_params:, creator_id:, token:)
...
           response[:school].save!
 
-          onboarded = SchoolOnboardingService.new(response[:school]).onboard(token:)
-          raise 'School onboarding failed' unless onboarded
+          SchoolOnboardingService.new(response[:school]).onboard(token:)
         end
 
         response
+      rescue ProfileApiClient::UnauthorizedError => e
+        # Do not log noise to sentry. The Profile API is only available to verified users over 13.
```
zetter-rpf: You've added 'The Profile API is only available to verified users over 13' to this comment. What does this mean - that all the instances from this error are from users under 13? How do we know that?

zetter-rpf (approved):
Great simplification, I've asked a clarifying question about a comment

lib/concepts/school/operations/create.rb
jamiebenstead: @zetter-rpf  You're right, we don't know that, so have removed it now. Claude lifted the comment from `profile_api_client.rb:264`

PR #973, opened 2026-08-19
Support service-authenticated Experience CS project and asset syncs

abcampo-iry:
## Status

- Related to https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1658
- https://github.com/RaspberryPiFoundation/experience-cs/pull/2344

### Security

- Creates a new key only used for assets upload and sync.
- Restricts the service identity to public Scratch projects, rejects user-owned, school-owned, and non-Scratch projects and limits asset uploads to public Code Classroom projects.

### Performance

- Checks stored checksums without downloading assets.
- Serializes writes to prevent conflicting or duplicate uploads.

## What’s changed?

- Added EXPERIENCE_CS_API_KEY configuration.
- Added authentication through the X-Experience-CS-API-Key header.
- Allowed Experience CS to create and update public Scratch projects.
- Allowed global asset uploads for public Code Classroom projects.
- Made asset uploads idempotent.

- ⚠️ Configure EXPERIENCE_CS_API_KEY with a secure secret.
- Deploy Editor API before enabling the Experience CS synchronization.
- Configure the same value as EDITOR_API_SYNC_API_KEY in Experience CS.

### DIFF
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -4,6 +4,7 @@ require 'project_loader'
 
 module Api
   class ProjectsController < ApiController
+    prepend_before_action :load_experience_cs_service_user, only: %i[create update]
     before_action :authorize_user, only: %i[create update index destroy]
     before_action :load_project, only: %i[show update destroy show_context]
     before_action :load_projects, only: %i[index]
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -4,8 +4,6 @@ require 'project_loader'
 
 module Api
   class ProjectsController < ApiController
-    EXPERIENCE_CS_SERVICE_PROJECT_TYPES = [Project::Types::SCRATCH, Project::Types::CODE_EDITOR_SCRATCH].freeze
-
     prepend_before_action :load_experience_cs_service_user, only: %i[create update]
     before_action :authorize_user, only: %i[create update index destroy]
     before_action :load_project, only: %i[show update destroy show_context]
@@ -68,34 +66,24 @@ module Api
     private
 
     def authorize_experience_cs_service_project
-      return unless current_user&.id == ExperienceCsServiceAuthenticator::USER_ID
-      return if experience_cs_service_project?
+      return unless current_user&.experience_cs_service_account?
+      return if experience_cs_service_project_change_permitted?
 
       raise CanCan::AccessDenied
     end
 
-    def experience_cs_service_project?
-      return public_scratch_project_attributes?(experience_cs_service_project_attributes) if action_name == 'create'
+    def experience_cs_service_project_change_permitted?
+      return false if action_name == 'update' && !@project.public_experience_cs_project?
 
-      public_scratch_project_attributes?(@project.attributes.symbolize_keys) &&
-        public_scratch_project_attributes?(experience_cs_service_project_attributes)
+      requested_experience_cs_project.public_experience_cs_project?
     end
 
-    def experience_cs_service_project_attributes
-      existing_attributes = if action_name == 'create'
-                              { user_id: nil, school_id: nil, project_type: Project.column_defaults['project_type'] }
-                            else
-                              @project.attributes.symbolize_keys
-                            end
-      requested_attributes = base_params.slice(:user_id, :school_id, :project_type).to_h.symbolize_keys
+    def requested_experience_cs_project
+      project = action_name == 'create' ? Project.new : @project.dup
+      attributes = base_params.slice(:user_id, :school_id, :project_type).to_h
+      project.assign_attributes(attributes)
 
-      existing_attributes.merge(requested_attributes)
-    end
-
-    def public_scratch_project_attributes?(attributes)
-      attributes[:user_id].nil? &&
-        attributes[:school_id].nil? &&
-        EXPERIENCE_CS_SERVICE_PROJECT_TYPES.include?(attributes[:project_type])
+      project
     end
 
     def verify_lesson_belongs_to_school
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -4,11 +4,14 @@ require 'project_loader'
 
 module Api
   class ProjectsController < ApiController
+    EXPERIENCE_CS_SERVICE_PROJECT_TYPES = [Project::Types::SCRATCH, Project::Types::CODE_EDITOR_SCRATCH].freeze
+
     prepend_before_action :load_experience_cs_service_user, only: %i[create update]
     before_action :authorize_user, only: %i[create update index destroy]
     before_action :load_project, only: %i[show update destroy show_context]
     before_action :load_projects, only: %i[index]
     load_and_authorize_resource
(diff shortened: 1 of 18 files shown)


app/services/experience_cs_service_authenticator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ExperienceCsServiceAuthenticator
+  HEADER = 'X-Experience-CS-API-Key'
+  USER_ID = '00000000-0000-0000-0000-000000000000'
+
+  def self.authenticate(candidate)
+    api_key = Rails.configuration.x.experience_cs.service_api_key
+    return if api_key.blank? || candidate.blank?
+    return unless ActiveSupport::SecurityUtils.secure_compare(candidate, api_key)
```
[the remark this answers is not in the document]
abcampo-iry: secure_compare already checks byte lengths and returns false when they differ. Only fixed_length_secure_compare raises. The existing unequal-length key spec confirms this, so the extra guard is redundant.

app/controllers/api/scratch/assets_controller.rb
```diff
@@ -52,11 +60,47 @@ def create_asset(attributes)
...
+        end
+      end
+
+      def attach_file_unless_present(scratch_asset, filename)
+        scratch_asset.file.attach(io: request.body, filename:) unless scratch_asset.file.attached?
+      end
+
+      def global_file_matches?(scratch_asset)
+        scratch_asset.file.blob.checksum == request_body_checksum
+      end
```
zetter-rpf, line 90: What's the reasoning behind needing to hash the assets? 

The asset name should already be the hash of the asset so I'm wondering if it's simpler to use that and skip saving it if the asset already exists.

Since global assets can only be created by trusted systems I think the risk is low (but I might be missing something?)

app/controllers/api/projects_controller.rb
```diff
@@ -64,6 +67,37 @@ def show_context
...
+
+      raise CanCan::AccessDenied
+    end
+
+    def experience_cs_service_project?
+      return public_scratch_project_attributes?(experience_cs_service_project_attributes) if action_name == 'create'
+
+      public_scratch_project_attributes?(@project.attributes.symbolize_keys) &&
+        public_scratch_project_attributes?(experience_cs_service_project_attributes)
+    end
```
zetter-rpf: I found this harder to follow what this was doing, I think it might be due to the naming of the `public_scratch_project_attributes?`

Could some of this be pushed into another object or the project model? Maybe a `.is_public_experience_cs_project?` or similar? I'm not sure if it it would work well as validation.

app/controllers/api/projects_controller.rb
```diff
@@ -63,6 +67,37 @@ def show_context
 
     private
 
+    def authorize_experience_cs_service_project
+      return unless current_user&.id == ExperienceCsServiceAuthenticator::USER_ID
```
zetter-rpf: This could be pushed onto the user to remove it from having to check the specific USER id  - `user.experience_cs_service_account?`

app/controllers/api/projects_controller.rb
```diff
@@ -63,6 +67,37 @@ def show_context
...
+      return if experience_cs_service_project?
+
+      raise CanCan::AccessDenied
+    end
+
+    def experience_cs_service_project?
```
zetter-rpf: Small naming thing - is the `service` part important here - might be easier to read if it was just `experience_cs_project?` unless it's specifically about the service account?

zetter-rpf:
I like the approach, and I think creating the 'experience cs user' is a good way to make it work with the existing controllers. 

I've added suggestions for improvements - let me know if you want to talk through any and I'll re-review after any changes since I want to be able to check the authorization logic again before mergin.

app/controllers/api/scratch/assets_controller.rb
abcampo-iry: For me the key is to compare checksums to avoid silently using the wrong asset if different content is uploaded with the same filename. It's quite inexpensive that check but you are right risk is low

app/controllers/api/projects_controller.rb
abcampo-iry: I moved this logic to Project.public_experience_cs_project?
I don’t think it should be a validation because the project can be valid in other contexts; this rule only limits what Experience CS can sync. Maybe if later we decide to be more global we can move it

app/controllers/api/projects_controller.rb
abcampo-iry: I added User.experience_cs_service_account?

app/controllers/api/projects_controller.rb
abcampo-iry: It applies specifically to Experience CS syncs, 
renamed it to experience_cs_service_project_change_permitted?
we can use cs_syncs if you think is more clear

app/controllers/api/scratch/assets_controller.rb
zetter-rpf: If it's reliable that ok - a reason to remove it would be if we didn't think it could hash files consistently because some part of them changed even though Scratch think it's the same file.

app/controllers/api/scratch/assets_controller.rb
```diff
@@ -24,30 +27,83 @@ def show
...
+        create_asset(
           project: @project_from_header,
           uploaded_user_id: current_user.id,
           filename: filename_with_extension
         )
+      end
+
+      def create_global
+        authorize! :create, @project_from_header
+        raise CanCan::AccessDenied unless experience_cs_public_project?
```
zetter-rpf: Is this check useful? If we're not saving a asset against a project, what's the reason in authorizing the project and checking that it's an experience CS one?

abcampo-iry: Good point. We should authorize the global asset upload, not project creation. We still check the project to ensure the asset is being uploaded for a valid Experience CS project.

zetter-rpf: > We should authorize the global asset upload

Agreed, I think it's allowed if the user is the experience CS admin using the shared secret?

>  We still check the project to ensure the asset is being uploaded for a valid Experience CS project.

This I'm still not sure about - if the asset isn't linked to a project, then I don't see any point in sending the project details or checking them. I could see the argument that the asset should be linked to a project, which we could do, we would just need to make sure that assets on public projects were still visible to all.

app/controllers/api/scratch/assets_controller.rb
```diff
@@ -24,30 +27,79 @@ def show
...
+        create_asset(
           project: @project_from_header,
           uploaded_user_id: current_user.id,
           filename: filename_with_extension
         )
+      end
+
+      def create_global
+        authorize! :create_global, ScratchAsset
+        raise CanCan::AccessDenied unless @project_from_header.public_experience_cs_project?
```
zetter-rpf: I talked to @abcampo-iry about this we agreed that there checking the project isn't useful here so can be removed. I was worried it might create a false sense of security, and removing it lets us slightly simplify the api call. This project check would be useful if we were saving the asset on the project.

zetter-rpf (approved):

PR #974, opened 2026-08-19
Allow students to view the preview projects

DNR500:
Needs to allow student view checks on issue: [1659](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1659)

Comes out of this work on editor-standalone - https://github.com/RaspberryPiFoundation/editor-standalone/pull/1064

## Summary

Updated `editor-api` CanCan permissions so that **school students** can `:show` the Experience CS preview starter project used by the classroom preview route.

With this changes students are now allowed to view **unowned** (`user_id: nil`, `school_id: nil`) **Scratch** template projects.

This prevents the preview route in editor-standalone from failing project loading (which previously triggered a redirect to `/${locale}/error`).

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -4,7 +4,7 @@ class Ability
   include CanCan::Ability
 
   def initialize(user)
-    define_common_non_student_abilities(user)
+    define_common_abilities
 
     return unless user
 
@@ -22,14 +22,7 @@ class Ability
 
   private
 
-  def define_common_non_student_abilities(user)
-    if user&.student?
-      # Allow students to view the Experience CS preview starter template.
-      can :show, Project, user_id: nil, school_id: nil, project_type: Project::Types::CODE_EDITOR_SCRATCH
-      can :show, Component, project: { user_id: nil, school_id: nil, project_type: Project::Types::CODE_EDITOR_SCRATCH }
-      return
-    end
-
+  def define_common_abilities
     # Anyone can view projects not owned by a user or a school.
     can :show, Project, user_id: nil, school_id: nil
     can :show, Component, project: { user_id: nil, school_id: nil }
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -23,7 +23,12 @@ class Ability
   private
 
   def define_common_non_student_abilities(user)
-    return if user&.student?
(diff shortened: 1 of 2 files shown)


app/models/ability.rb
```diff
@@ -23,7 +23,12 @@ def initialize(user)
   private
 
   def define_common_non_student_abilities(user)
-    return if user&.student?
+    if user&.student?
+      # Allow students to view the Experience CS preview starter template.
+      can :show, Project, user_id: nil, school_id: nil, project_type: Project::Types::CODE_EDITOR_SCRATCH
+      can :show, Component, project: { user_id: nil, school_id: nil, project_type: Project::Types::CODE_EDITOR_SCRATCH }
+      return
```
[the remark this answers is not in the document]
DNR500: Addressed

app/models/ability.rb
```diff
@@ -23,7 +23,12 @@ def initialize(user)
...
+    end
```
zetter-rpf: I think the placement of these rules aren't quite right because they are in the `non_student_abilities` method but apply to students.

If these projects should be visible to everyone (which I think is the case), then they should be here without the `if user&.student?` check

If these projects should be visible by student (but not the general public) then they should live in a `define_school_student_abilities` method, or a more general `define_school_student_global_abilities` method since these abilities aren't specific to a school like the others are.

zetter-rpf: I've just seen the rules below which I missed before. Is it important we restrict students from seeing other public projects (which they can presumably see anyway if they sign out?). If not, maybe this could be solved by removing the `return if user&.student?` return and renaming `define_common_non_student_abilities` to `define_common_abilities`

DNR500: Removed the student early-return and Scratch-only exception, and renamed `define_common_non_student_abilities` to `define_common_abilities`, so students can view unowned public projects the same as everyone else.

Here’s the [commit](https://github.com/RaspberryPiFoundation/editor-api/pull/974/changes/49ab7fe7ae9e28540dc80d13f7f2d0a50e4908bf) - Think that’s what you meant?

zetter-rpf (approved):
That is what I meant, and I think that's clearer.

PR #975, opened 2026-08-20
Identify Experience CS projects

cocomarine:
## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1717

## Points for consideration:

- Security
- Performance

## What's changed?
- Add `origin` column to Project
- Validations: 
  - `inclusion` in known origins (only `Project::Origins::EXPERIENCE_CS` for now) 
  - allows nil value so that we can backfill, blocks any change once set
- build_project now sets origin for ECS admins
- Remix projects inherit `origin` from original project. 
- With #973 in, service-authenticated create also gets the origin assignment.

### DIFF
--- a/db/migrate/20260820122510_add_origin_to_projects.rb
+++ b/db/migrate/20260820122510_add_origin_to_projects.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddOriginToProjects < ActiveRecord::Migration[8.1]
+  def change
+    add_column :projects, :origin, :string
+    add_index :projects, :origin, where: 'origin IS NOT NULL'
+  end
+end
--- a/db/migrate/20260820122510_add_origin_to_projects.rb
+++ b/db/migrate/20260820122510_add_origin_to_projects.rb
@@ -1,12 +1,7 @@
 # frozen_string_literal: true
 
 class AddOriginToProjects < ActiveRecord::Migration[8.1]
-  disable_ddl_transaction!
-
   def change
-    add_column :projects, :origin, :string, if_not_exists: true
-    add_index :projects, :origin, where: 'origin IS NOT NULL',
-                                  algorithm: :concurrently,
-                                  if_not_exists: true
+    add_column :projects, :origin, :string
   end
 end
--- a/db/migrate/20260820122510_add_origin_to_projects.rb
+++ b/db/migrate/20260820122510_add_origin_to_projects.rb
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 
 class AddOriginToProjects < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
   def change
-    add_column :projects, :origin, :string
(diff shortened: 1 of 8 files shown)


lib/concepts/project/operations/create.rb
```diff
@@ -17,7 +17,12 @@ def call(project_hash:, current_user:)
       private
 
       def build_project(project_hash, current_user)
-        project_hash[:identifier] = PhraseIdentifier.generate unless current_user&.experience_cs_admin?
+        if current_user&.experience_cs_admin?
+          project_hash[:origin] = Project::Origins::EXPERIENCE_CS
+        else
+          project_hash[:identifier] = PhraseIdentifier.generate
+        end
+
```
zetter-rpf, line 25: Are there any the tradeoffs for setting the project origin here compared to getting the caller  (experience CS) to set it and passing in the param?

cocomarine: I set it this way so that it is something fully server-controlled and cannot be changed easily. This way, we have tighter control on them. If we were to support a lot more origins, perhaps passing it as a param make sense as we don't have to update whenever new origin is added.

db/migrate/20260820122510_add_origin_to_projects.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddOriginToProjects < ActiveRecord::Migration[8.1]
+  def change
+    add_column :projects, :origin, :string
+    add_index :projects, :origin, where: 'origin IS NOT NULL'
+  end
+end
```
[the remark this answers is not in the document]
cocomarine: added

db/migrate/20260820122510_add_origin_to_projects.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddOriginToProjects < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  def change
+    add_column :projects, :origin, :string, if_not_exists: true
+    add_index :projects, :origin, where: 'origin IS NOT NULL',
+                                  algorithm: :concurrently,
+                                  if_not_exists: true
```
zetter-rpf: Is there a reason that this needs an index?

Normally you would only add indexes on things you would need to look up a record (often identifiers and foreign keys). If we're never looking for all the records with a particular origin, maybe this isn't needed?

Counter-intuitively, adding more indexes than you need can slow your database down as it makes writing to your database slower and can use up more storage space.

zetter-rpf (approved):
Nice one

One suggestion to remove the index if it's not needed.

db/migrate/20260820122510_add_origin_to_projects.rb
cocomarine: Good point. Added it in case we need it later, but no need now. Removed.

PR #979, opened 2026-08-25
Support syncing ExperienceCS user projects

abcampo-iry:
## Status

- Closes partially https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1709
- Related to https://github.com/RaspberryPiFoundation/experience-cs/pull/2357

## What's changed?

Adds an endpoint for migrating an ExperienceCS project into its existing Code Classroom stub.

The migration:

- Preserves the project identifier, locale, owner and permissions.
- Copies the name, instructions, Scratch blocks and assets.
- Keeps uploaded assets private.
- Supports safe retries.
- Uses service authentication instead of the project creator’s token.

### DIFF
--- a/app/controllers/api/experience_cs_project_migrations_controller.rb
+++ b/app/controllers/api/experience_cs_project_migrations_controller.rb
@@ -5,7 +5,6 @@ module Api
     prepend_before_action :load_experience_cs_service_user
     before_action :authorize_user
     before_action :load_project
-    before_action :authorize_migration
 
     def update
       migrate_project!
@@ -24,17 +23,13 @@ module Api
       @project = Project.find_by!(identifier: params.expect(:id), locale: nil)
     end
 
-    def authorize_migration
-      authorize! :migrate_from_experience_cs, @project
-    end
-
     def migrate_project!
       attributes = migration_params
       @project.with_lock do
+        authorize! :migrate_from_experience_cs, @project
         @project.update!(
           attributes.slice(:name, :instructions).merge(
-            project_type: Project::Types::CODE_EDITOR_SCRATCH,
-            experience_cs_migrated_at: @project.experience_cs_migrated_at || Time.current
+            project_type: Project::Types::CODE_EDITOR_SCRATCH
           )
         )
         scratch_component = @project.scratch_component || @project.build_scratch_component
--- a/app/controllers/api/experience_cs_project_migrations_controller.rb
+++ b/app/controllers/api/experience_cs_project_migrations_controller.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Api
+  class ExperienceCsProjectMigrationsController < ApiController
+    prepend_before_action :load_experience_cs_service_user
+    before_action :authorize_user
+    before_action :load_project
+    before_action :authorize_migration
+
+    def update
+      migrate_project!
+      render json: {
+        identifier: @project.identifier,
+        locale: @project.locale,
+        project_type: @project.project_type
+      }
+    rescue ActiveRecord::RecordInvalid, ActionController::ParameterMissing => e
+      render json: { error: e.message }, status: :unprocessable_content
+    end
+
+    private
+
+    def load_project
+      @project = Project.find_by!(identifier: params.expect(:id), locale: nil)
+    end
+
+    def authorize_migration
+      authorize! :migrate_from_experience_cs, @project
+    end
+
+    def migrate_project!
+      attributes = migration_params
+      @project.with_lock do
+        @project.update!(
+          attributes.slice(:name, :instructions).merge(
+            project_type: Project::Types::CODE_EDITOR_SCRATCH,
+            experience_cs_migrated_at: @project.experience_cs_migrated_at || Time.current
+          )
+        )
+        scratch_component = @project.scratch_component || @project.build_scratch_component
+        scratch_component.update!(attributes.require(:scratch_component).slice(:content))
+      end
+    end
+
+    def migration_params
+      params.fetch(:project, {}).permit(
+        :name,
+        { instructions: [:markdown_content] },
+        { scratch_component: { content: {} } }
+      )
+    end
+  end
+end
--- a/app/controllers/api/scratch/assets_controller.rb
+++ b/app/controllers/api/scratch/assets_controller.rb
@@ -7,10 +7,10 @@ module Api
     class AssetsController < ApiController
       include ActiveStorage::SetCurrent
 
-      prepend_before_action :load_experience_cs_service_user, only: %i[create_global]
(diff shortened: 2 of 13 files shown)


app/models/project.rb
```diff
@@ -108,6 +108,12 @@ def public_experience_cs_project?
     user_id.nil? && school_id.nil? && EXPERIENCE_CS_PROJECT_TYPES.include?(project_type)
   end
 
+  def experience_cs_migration_target?
+    return false unless user_id.present? && school_id.present?
+
+    project_type == Types::SCRATCH || experience_cs_migrated_at.present?
+  end
```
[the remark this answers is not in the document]
abcampo-iry: I’ve restricted migration targets to Experience CS Scratch projects and previously migrated Code Classroom Scratch projects.

app/controllers/api/scratch/assets_controller.rb
```diff
@@ -100,14 +102,46 @@ def request_body_checksum
...
-          identifier:,
-          project_type: Project::Types::CODE_EDITOR_SCRATCH
-        )
+        project_scope = Project.where(identifier:)
+        project_scope = project_scope.where(locale: nil) if current_user&.experience_cs_service_account?
+        @project_from_header = project_scope.first!
+        return if project_accepts_asset_upload?(@project_from_header)
+
+        raise ActiveRecord::RecordNotFound, 'Not Found'
+      end
```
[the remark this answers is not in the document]
abcampo-iry: I’ve constrained the lookup to the appropriate Scratch project types before selecting the project.

app/models/ability.rb
```diff
@@ -72,6 +73,12 @@ def define_school_owner_abilities(school:)
     can(%i[read], :school_member)
     can(%i[read create import update destroy regenerate_join_code], SchoolClass, school: { id: school.id })
     can(%i[read update show_context], Project, school_id: school.id, lesson: { visibility: %w[teachers students] })
+    lesson_project_ids = Project.where(
+      school_id: school.id,
+      remixed_from_id: nil,
+      lesson_id: Lesson.where(school_id: school.id, visibility: %w[teachers students]).select(:id)
+    ).pluck(:id)
+    can(%i[read show_context], Project, school_id: school.id, remixed_from_id: lesson_project_ids)
```
zetter-rpf: Could you explain why this extra permission is needed? I can't see why the `can(%i[read update show_context], Project` permission above isn't sufficient and why this is important for School owners but not teachers.


If we do need it, I wonder if there's a better way to write it - currently it will do two queries on every page load for project owners even if the permission isn't used. Maybe Scopes can be used to simplify or [cancancan's block mode](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/define_abilities_with_blocks.md) can be used

app/controllers/api/experience_cs_project_migrations_controller.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Api
+  class ExperienceCsProjectMigrationsController < ApiController
```
zetter-rpf, line 4: Handling this in a new controller is a good idea and makes it easier to remove it when we have done the migration.

app/models/project.rb
```diff
@@ -108,6 +108,13 @@ def public_experience_cs_project?
     user_id.nil? && school_id.nil? && EXPERIENCE_CS_PROJECT_TYPES.include?(project_type)
   end
 
+  def experience_cs_migration_target?
+    return false unless user_id.present? && school_id.present?
+
+    project_type == Types::SCRATCH ||
+      (project_type == Types::CODE_EDITOR_SCRATCH && experience_cs_migrated_at.present?)
```
zetter-rpf: Could you explain the reasoning for introducing the `experience_cs_migrated_at` column? 

Re-running the migration is good could be useful, however it could also cause data to be lost if updates have been made in Code Classroom after a project has been migrated.

I wonder if it would be simpler to just allow `Types::SCRATCH` to be migrated. If we do need to fix anything in the migration, we'll have to consider what to do when it comes up.

app/controllers/api/scratch/assets_controller.rb
```diff
@@ -7,10 +7,10 @@ module Scratch
     class AssetsController < ApiController
       include ActiveStorage::SetCurrent
 
-      prepend_before_action :load_experience_cs_service_user, only: %i[create_global]
+      prepend_before_action :load_experience_cs_service_user, only: :create_global
+      prepend_before_action :load_project_asset_context, only: %i[show create]
       before_action :authorize_user, except: %i[show]
-      prepend_before_action :load_project_from_header, only: %i[show create]
-      authorize_resource :project_from_header, except: %i[create_global]
+      before_action :authorize_project_from_header, except: %i[create_global]
```
zetter-rpf: I am finding this much harder to follow the loading and authorisation flows with these changes so have less confidence everything is authorized as expected.

Since this is only temporarily needed, I wonder if having a seperate action for adding assets to a project for the ExCS user could simplify this and reduce the risk of getting auth wrong. It would also mean it would be easier to remove the functionality once the migration is complete.

app/models/ability.rb
abcampo-iry: Student projects only link to lessons through their parent project, so this extra rule is needed; I'm rewritting it with your proposal to avoid extra queries.

app/models/project.rb
abcampo-iry: I added it to identify previously migrated projects and allow retries, but you’re right that retrying could overwrite newer Code Classroom  changes, i'm removing the column

app/controllers/api/scratch/assets_controller.rb
abcampo-iry: Agreed, i'm going to move to its own action so it's simpler. Thanks for the comments, they do make sense, i was really focus on experience-cs to work.

app/models/ability.rb
zetter-rpf: I still am a bit confused. 

What endpoint or call to authorize won't work as needed if this doesn't exist?

Doesn't the more general permission above do everything we need? (which is now very similar except it doesn't check the parent relationship)

abcampo-iry: /api/projects/:id could return 403 when a school owner who is not a class teacher opens a migrated student project.
A student project has no lesson_id; it reaches the lesson through its parent.
We only need this for the case where school owners must be able to view the migrated project.

this way we follow the permissions in the ticket -> the student, teachers in the class, owners can see the project

If “owners” does not mean school owners, or that access is not required, we can remove it.

zetter-rpf: Thanks for explaining. I could be a problem for existing flows which I hadn't realized from the change.

zetter-rpf (approved):
Nice, thanks for changing the assets controller, I found that a lot easier to understand

PR #980, opened 2026-08-25
Return the project locale on lesson JSON

jamiebenstead:
## Status

- Work towards - https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1722

Return the project locale on lesson JSON

Adds :locale to the nested project object in _lesson.json.jbuilder, so it appears on GET /api/lessons, GET /api/lessons/:id and POST /api/lessons/batch.

Experience CS now sends project_attributes.locale when adding projects to a class, so teachers get lesson content in the language they were browsing in. Editor API already accepted and stored that value, but never returned it — so there's no way to confirm through the API which locale a lesson's project was created in.

This follows the existing pattern on this endpoint: create_batch already echoes origin_identifier back so callers can verify their request landed correctly. It also brings the nested project in line with projects#show, which already returns locale.

Response-only — no change to what gets persisted, and no behaviour change for existing consumers (additive field). No caller reads it today; this is for consistency and observability rather than to unblock the Experience CS work, which ships without it.

### DIFF
--- a/app/views/api/lessons/_lesson.json.jbuilder
+++ b/app/views/api/lessons/_lesson.json.jbuilder
@@ -19,7 +19,8 @@ if lesson.project
   json.project(
     lesson.project,
     :identifier,
-    :project_type
+    :project_type,
+    :locale
   )
 end
 
--- a/spec/features/lesson/creating_a_batch_of_lessons_spec.rb
+++ b/spec/features/lesson/creating_a_batch_of_lessons_spec.rb
@@ -75,6 +75,33 @@ RSpec.describe 'Creating a batch of lessons', type: :request do
     end
   end
 
+  context 'when a locale is supplied for each project' do
+    let(:lesson_project_params) do
+      [
+        {
+          name: 'Lesson 1',
+          school_id: school.id,
+          project_attributes: { name: 'Project 1', project_type: Project::Types::CODE_EDITOR_SCRATCH, locale: 'fr-FR' }
+        },
+        {
+          name: 'Lesson 2',
+          school_id: school.id,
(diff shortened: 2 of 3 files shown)


zetter-rpf:
@jamiebenstead could you add a PR description for this?

I'm not sure where this is needed - where are we calling the lesson listing API that we will need the locale?

EDIT- this is clearer after I looked at https://github.com/RaspberryPiFoundation/experience-cs/pull/2358 and saw it being used, but it would be still good to explain that in the description.

jamiebenstead:
> @jamiebenstead could you add a PR description for this?
> 
> I'm not sure where this is needed - where are we calling the lesson listing API that we will need the locale?
> 
> EDIT- this is clearer after I looked at [RaspberryPiFoundation/experience-cs#2358](https://github.com/RaspberryPiFoundation/experience-cs/pull/2358) and saw it being used, but it would be still good to explain that in the description.

@zetter-rpf Added a description now. It seems like we don't actually need to make this change, but Claude recommends it based on the [experience-cs](https://github.com/RaspberryPiFoundation/experience-cs/pull/2358) work

zetter-rpf:
> @zetter-rpf Added a description now. It seems like we don't actually need to make this change,

Agreed. There might be some API changes needed for https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1724, but don't think we'll need to set the Project locale.

zetter-rpf:
> @zetter-rpf Added a description now. It seems like we don't actually need to make this change,

Agreed. There might be some API changes needed for https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1724, but don't think we'll need to set the Project locale.

PR #982, opened 2026-08-25
Add ExCS projects

cocomarine:
## Status

- Related to https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1724

## What's changed?

- Teachers can create a lesson from an existing Experience CS project (RpfProject)
- POST /api/lessons accepts a new param: lesson[source_project_identifier].
- Lesson::Create builds the lesson with a copy of the source project if it's provided.
- Refactored project copying into a module lib/concepts/project/copying.rb 
  - Now used by Lesson::Create, Lesson::CreateCopy, Project::CreateRemix.

## Follow-ups (not in this PR)

- Decide whether the `remixed_from_id` is used to track the source project or, use another field for it. 
- Adding projects in bulk (using the UnitClassroomAssigner)
- ❓ Before this PR, CreateCopy was only copying images only, not video/audio. Is that OK?


lib/concepts/lesson/operations/create.rb
```diff
@@ -21,14 +22,30 @@ def call(lesson_params:)
...
+
+      def remix_attributes(source_project, project_params)
+        {
+          locale: nil,
+          name: project_params[:name].presence || source_project.name,
+          user_id: project_params[:user_id],
+          school_id: project_params[:school_id],
+          lesson_id: project_params[:lesson_id]
+        }
+      end
```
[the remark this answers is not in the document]
cocomarine: This will be worked on a separate issue as this is not as simple as just adding those attributes and more thoughts are required in terms of teacher project logic and ability.

lib/concepts/lesson/operations/create.rb
```diff
@@ -21,14 +22,30 @@ def call(lesson_params:)
...
-                                                                  lesson_id: new_lesson.id })
-        new_lesson.project = Project.new(project_params)
+        project_params = (lesson_hash[:project_attributes] || {}).merge({ user_id: lesson_hash[:user_id],
+                                                                          school_id: lesson_hash[:school_id],
+                                                                          lesson_id: new_lesson.id })
+        new_lesson.project = source_project ? build_remix(source_project, project_params) : Project.new(project_params)
         new_lesson
       end
+
+      def build_remix(source_project, project_params)
```
zetter-rpf: Maybe this should be `build_copy` or `copy_project`?

I think it's a good idea to call this Copying if we think it is a different action from Remixing, but it would be good to be consistent with the names.

zetter-rpf (approved):
Nice one, just left a suggestion about switching between calling this a remix vs a copy

lib/concepts/lesson/operations/create.rb
```diff
@@ -21,14 +22,30 @@ def call(lesson_params:)
...
+        Project::Copying.copy_project(source_project,
+                                      attributes: remix_attributes(source_project, project_params),
+                                      media: %i[images videos audio])
+      end
+
+      def remix_attributes(source_project, project_params)
```
zetter-rpf: Similar to the `build_remix` note above

PR #986, opened 2026-08-26
Reject unassociated teacher accounts in Join Code flow

mwtrew:
Also adds tests, updates the user factory to be aware of the new sub field, and updates some existing tests to take advantage of the factory.

## Status

- Related to https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1595

## What's changed?
- Uses the `sub` field from Hydra to prevent "full" (i.e. expected to be a teacher) accounts from erroneously going down the student add-to-class flow.

### DIFF
--- a/app/services/join_status_service.rb
+++ b/app/services/join_status_service.rb
@@ -38,7 +38,7 @@ class JoinStatusService
 
   # The user has no role in this school yet: may they join as a new student?
   def new_user_join_status
-    return :not_a_student if user_has_non_student_role? || user_has_non_student_account_type?
+    return :not_a_student if user_has_non_student_account_type?
     return :wrong_school if user_in_different_school?
     return :domain_mismatch unless @school.email_domain_in_school_domains?(@user.email)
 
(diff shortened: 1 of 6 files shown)


app/services/join_status_service.rb
```diff
@@ -38,7 +38,7 @@ def existing_user_join_status
 
   # The user has no role in this school yet: may they join as a new student?
   def new_user_join_status
-    return :not_a_student if user_has_non_student_role?
+    return :not_a_student if user_has_non_student_role? || user_has_non_student_account_type?
```
peconia:
```suggestion
    return user_has_non_student_account_type?
```

Should this now be enough for the check?

mwtrew: Seems to affect another test case, which I'd rather not dive into just now:

```
     Failure/Error: expect(data[:status]).to eq('not_a_student')
     
       expected: "not_a_student"
            got: "wrong_school"
     
       (compared using ==)
     # ./spec/requests/join_controller_spec.rb:108'
```

zetter-rpf (approved):

app/services/join_status_service.rb
mwtrew: Actually it was an existing problem with that test suite, I've fixed it and your suggestion works.

PR #987, opened 2026-08-26
Seed different project types

zetter-rpf:
## Status

- Related to test PR: https://github.com/RaspberryPiFoundation/learner-experience-integration-tests/pull/72
- Related to issue: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1749

## What's changed?

Previously we only created Python projects in our seed data which made integration testing harder. 
This PR also creates Scratch and HTML projects and gives them clearer names.

I've also turned the Scratch feature flag on for the School to make the setup for the Scratch tests simpler.

I've also changed the local dev seeds to use the same method, so it will be easier to test different projects in dev too.

After this is deployed, https://github.com/RaspberryPiFoundation/learner-experience-integration-tests/pull/72
should be merged immediately after 

*Made with help from Claude*


spec/lib/test_seeds_spec.rb
```diff
@@ -164,12 +164,20 @@
...
-      expect(Project.where(lesson_id: lesson.pluck(:id)).length).to eq(4)
+      expect(lesson.length).to eq(6)
+      expect(lesson.pluck(:name)).to contain_exactly(
+        'Lesson 1 python', 'Lesson 2 html/css', 'Lesson 3 scratch',
+        'Lesson 1 python', 'Lesson 2 html/css', 'Lesson 3 scratch'
+      )
+
+      projects = Project.where(lesson_id: lesson.pluck(:id))
+      expect(projects.length).to eq(6)
+      expect(projects.find_by(name: 'Lesson 2 html/css').project_type).to eq(Project::Types::HTML)
```
DNR500, line 179: These only check one HTML project and one Scratch project, but we create two of each (one per class). They also don’t check that the project content is actually there, or that Scratch is enabled on the school. Should we cover those in the specs?

lib/tasks/seed_data/lesson_scratch_starter.json
```diff
@@ -0,0 +1,98 @@
+{
```
DNR500, line 1: This looks very similiar to `excs_preview_starter.json` (maybe only the formatting differs). Should we reuse that file instead of adding another copy?

DNR500 (approved):
Looks good, nice to have this to better support those test in LXIT

lib/tasks/seed_data/lesson_scratch_starter.json
zetter-rpf: Good spot, I've combined them

spec/lib/test_seeds_spec.rb
zetter-rpf: Agreed, I've added some more test cases and unit test coverage for some of the seed helpers.

PR #989, opened 2026-08-28
Record source project when adding ExCS projects

cocomarine:
## Status

- Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1724

## Why not reuse `remixed_from_id` ?

- Although they are created by similar processes, ExCS projects added to editor-api are different from remixes. 
- A lesson project copied from an ExCS source would have a non-nil remixed_from_id -> can't be teacher_project_ids with current logic in ability -> regression in teacher's ability such as teachers losing read/show_context on their students' remixes of that lesson etc. 
- Api::Projects::RemixesController#index and RemixSelection both query Project.where(remixed_from_id: <id>), so the ExCS source project would start reporting lesson projects as its remixes, which is not correct. 
- **_For these reasons, decided to go with new column specifically for recording source_project._**  

## What's changed?

- Added `source_project_id` column
- Updated attributes_for_copy of lesson create.rb to add source project
  - Also cleared remix columns for above reasons (and for the unlikely situation where source projects have these columns populated)
- Added tests for asserting source_projects are copied during lesson copy and project remix.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -18,6 +18,8 @@ class Project < ApplicationRecord
   belongs_to :lesson, optional: true
   belongs_to :parent, optional: true, class_name: :Project, foreign_key: :remixed_from_id, inverse_of: :remixes
   has_many :remixes, dependent: :nullify, class_name: :Project, foreign_key: :remixed_from_id, inverse_of: :parent
+  belongs_to :source_project, optional: true, class_name: :Project, inverse_of: :derived_projects
+  has_many :derived_projects, dependent: :nullify, class_name: :Project, foreign_key: :source_project_id, inverse_of: :source_project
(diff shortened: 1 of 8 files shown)


zetter-rpf:
Thanks for investigating. I want to think about this a bit more - I'm worried adding another similar column to projects might be confusing so want to think through if there any any alternatives or ways to make this clearer.

zetter-rpf (approved):
Great,

I've had a longer think about this, the thing I don't like about the existing code is how the permissions rely on remixed_from_id to determine what students and teachers should do with a project. I think if a project has been remixed should be a separate concern to if a student or teacher should be able to see it. Currently checking remixes does add complexity to the abilities as we're doing additional database loads for this.

But, even if we changed the abilities, I don't think it would be right to populate the remixed_from_id as part of this - remixing is a different action to what we're doing here (as you said in the description).

PR #1002, opened 2026-09-07
Add blocks project seed for seeding

jamdelion:
## What's changed?

_PR description written by Claude, and reviewed by JH_

Adds a Blocks (Scratch) starter project to the seed data, so developers working on integrated
Blocks editor projects in projects-ui have a real example locally instead of having to hand-craft
a `scratch_component` row.

- New `lib/tasks/project_components/neil-the-seal-starter/` containing `neil-the-seal-starter.sb3`
  and a `project_config.yml` with `TYPE: "code_editor_scratch"`. Picked up automatically by
  `rails projects:create_all`, which `db/seeds.rb` already invokes.
- The `.sb3` is copied unmodified from
  [`raspberrypilearning/editor-neil-the-seal`](https://github.com/raspberrypilearning/editor-neil-the-seal)
  at `en/code/neil-the-seal-starter/neil-the-seal-starter.sb3`. The identifier
  `neil-the-seal-starter` matches production, so local and deployed data agree.
- Documents Scratch seeds in `lib/tasks/README.md`, which previously only covered Python and HTML.
  Most usefully it warns that content repositories use a **lowercase-keyed** `project_config.yml`
  read by the webhook/`UploadJob` path, whereas the seeding task reads **uppercase** keys — copying
  one across verbatim parses fine but silently produces a nameless Python project.
- Relaxes the README's claim that `IDENTIFIER` is "three words separated by dashes"; it isn't
  enforced and existing projects such as `editor-scratch-testing-starter` already use more.

### Related

Reaching this project through projects-ui also needs the projects-admin seed that sets
`direct_to_editor` and `editor_starter_project` — this happens using the `production-to-local` script. Once both are
seeded it's at `/en/projects/editor-neil-the-seal/editor`. It's unlisted and in no pathway,
matching the current status of production, so it can't be reached by browsing.


zetter-rpf (approved):
Nice one.

I've added you to the CLA list so you should be able to merge once you rebase/merge in main.

PR #1011, opened 2026-09-11
Send owner transfer email when requested

cocomarine:
## Status

- Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1766

## What for?
- This PR adds backend functionality to send an email to a prospective owner (when the current school owner clicks transfer button). 
- Out of scope: 
  - Actual ownership transfer
  - Hooking it up to the frontend

## What's changed?

-  Added `SchoolOwnershipMailer` and template
   - Add instance variables (such as those for current owner name and prospective owner name) to the email template when controller actions are set up and plumbed to the frontend. 
   - Followed what the invitation emails do and added text.erb template (deviating slightly from copy doc due to lack of formatting inline link and bullet points)
-  Added `ownership_transfers` table and its model
- The model calls and queues the mailer and the mailer reads @school off the ownership_transfer param to build the email.

## Emails
- Preview at http://localhost:3009/rails/mailers :

Screenshot 2026-09-14 at 12 18 43

- When triggered in rails console:
  - Queues GoodJob
  Screenshot 2026-09-14 at 11 09 38
  - Email gets sent via Postmark dev sandbox
Screenshot 2026-09-14 at 13 15 43

### DIFF
--- a/app/mailers/school_ownership_mailer.rb
+++ b/app/mailers/school_ownership_mailer.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class SchoolOwnershipMailer < ApplicationMailer
+  default from: email_address_with_name('web@raspberrypi.org', 'Raspberry Pi Foundation')
+
+  def request_ownership_transfer
+    @school = params[:ownership_transfer].school
+    @token = params[:ownership_transfer].generate_token_for(:ownership_transfer)
+
+    mail(to: params[:ownership_transfer].email_address,
+         subject: "You have been asked to take ownership of #{@school.name}",
+         track_opens: 'true',
+         message_stream: 'outbound')
+  end
+end
--- a/app/models/ownership_transfer.rb
+++ b/app/models/ownership_transfer.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class OwnershipTransfer < ApplicationRecord
+  delegate :name, to: :school, prefix: true
+
+  belongs_to :school
+  validates :email_address,
+            format: { with: EmailValidator.regexp, message: I18n.t('validations.invitation.email_address') }
+  after_create_commit :send_ownership_transfer_request_email
+  encrypts :email_address
+
+  generates_token_for :ownership_transfer, expires_in: 30.days do
+    email_address
+  end
+
+  private
+
+  def send_ownership_transfer_request_email
+    SchoolOwnershipMailer.with(ownership_transfer: self).request_ownership_transfer.deliver_later
+  end
+end
--- a/.env.example
+++ b/.env.example
@@ -34,7 +34,7 @@ SMEE_TUNNEL=https://smee.io/MLq0n9kvAes2vydX
 # BYPASS_OAUTH=true
 
 HOST_URL=http://localhost:3009
-EDITOR_PUBLIC_URL=http://localhost:3012
+EDITOR_PUBLIC_URL=http://classroom.localhost:3013
 
 PROFILE_API_KEY=test # This has to match the value set in Profile (https://github.com/RaspberryPiFoundation/profile/blob/ca10a4f360b6fe2b04be76264e03283054126b0f/.env.example#L45).
 
--- a/app/views/school_ownership_mailer/request_ownership_transfer.text.erb
+++ b/app/views/school_ownership_mailer/request_ownership_transfer.text.erb
@@ -1,0 +1,14 @@
+You have been asked to take ownership of:
+
+<%= @school.name %>
+
+Log in to your account to accept this request and become the owner of this school account.
+
+Accept ownership transfer:
+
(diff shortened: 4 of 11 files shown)


app/mailers/school_ownership_mailer.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class SchoolOwnershipMailer < ApplicationMailer
+  default from: email_address_with_name('web@raspberrypi.org', 'Raspberry Pi Foundation')
+
+  def request_ownership_transfer
+    ownership_transfer = params[:ownership_transfer]
+    @school = ownership_transfer.school
+    @token = ownership_transfer.generate_token_for(:ownership_transfer)
```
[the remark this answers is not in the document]
zetter-rpf: I agree - if we expect users to already have an account sign in to accept a transfer, what is the purpose of creating a token?

app/views/school_ownership_mailer/request_ownership_transfer.html.erb
```diff
@@ -0,0 +1,21 @@
+<p>Hi <%= @nominee_name || 'there' %>,</p>
```
zetter-rpf: If we don't set @nominee_name anywhere it might be clearer to remove it.

app/views/school_ownership_mailer/request_ownership_transfer.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+<p><%= @current_owner_name || @school.name %> has nominated you to become the new owner of the Code Classroom account for <%= @school.name %>.</p>
```
zetter-rpf: Similar to above

zetter-rpf:
> ❓ Do we need both text.erb and html.erb templates?

I would follow what the invitation emails do unless there's a reason to change

app/views/school_ownership_mailer/request_ownership_transfer.html.erb
```diff
@@ -0,0 +1,21 @@
...
+<p>If you accept, you’ll take over ownership of the school account and be able to:</p>
+
+<ul>
+  <li>Invite a new teacher to the account</li>
+  <li>Manage teachers, students, and their access</li>
+  <li>View and manage Code Classroom settings</li>
+  <li>Transfer ownership to someone else in the future</li>
+</ul>
+
+<p>To continue, please <%= link_to 'log in', "#{ENV.fetch('EDITOR_PUBLIC_URL')}/en-US/school" %> to your account to review and accept the ownership transfer.</p>
```
zetter-rpf: Do we need a local-less redirect that we can link people to them without overriding the users preferred browser locale? Or save the users prefered locale on the server

Might not be part of this

zetter-rpf (commented):
I've added some suggestions of things that might be able to be removed as they are not being used/set up as part of this so keeping them might be confusing.

Let me know if I can re-review

app/mailers/school_ownership_mailer.rb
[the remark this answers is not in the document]
cocomarine: @token will be wired into the accept URL once the follow-up PRs, so removing for now.

app/views/school_ownership_mailer/request_ownership_transfer.html.erb
cocomarine: OK, changed to local-less path.

cocomarine:
> > ❓ Do we need both text.erb and html.erb templates?
> 
> I would follow what the invitation emails do unless there's a reason to change

OK, removed html.erb.

app/models/ownership_transfer.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  belongs_to :school
+  validates :email_address,
+            format: { with: EmailValidator.regexp, message: I18n.t('validations.invitation.email_address') }
+  after_create_commit :send_ownership_transfer_request_email
+  encrypts :email_address
+
+  generates_token_for :ownership_transfer, expires_in: 30.days do
+    email_address
+  end
```
zetter-rpf: Is this still needed?

app/mailers/school_ownership_mailer.rb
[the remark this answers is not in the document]
zetter-rpf: Is that written on the issue somewhere? My understanding was that we wouldn't let someone accept without them signing in?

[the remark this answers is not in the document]
zetter-rpf: Me and @cocomarine talked about this, we agreed that even though our existing invitation system uses tokens, we didn't need them for this as we'll verify they are being redeemed by the correct person based on what account they are logged in with.

We said that we might want a user_id/invited_user_id column on the table to help with this, but that doesn't need to be added as part of this PR.

zetter-rpf (approved):
Approved given that we'll remove the references to tokens

PR #1023, opened 2026-09-17
Handle RecordInvalid in concurrent asset upload scenario

mwtrew:
This arose when the first asset row is committed before validation on the second takes place.

## Status

- Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1785

## What's changed?

- Handles RecordInvalid where it was caused by a duplicate asset upload.


zetter-rpf:
Do you have any thoughts as to why this is happening - for example, is this caused by a project having the same asset twice? Or the same user saving a project multiple times in quick succession? Or something else?

Just trying to understand in case this points to odd behaviour in the editor uploading the same asset multiple times.

jamiebenstead (approved):

mwtrew:
@zetter-rpf I didn't look into why that might be happening from the client side - thought it was worth fixing API-side regardless, but might be worth looking into.

PR #1025, opened 2026-09-18
Accept or decline transfer of ownership

DNR500:
## Summary

- Backend part of issue [1768](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1768) 
- Tracked on issue [1814](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1814)
- Frontend work is on https://github.com/RaspberryPiFoundation/editor-standalone/pull/1109

This adds the two endpoints that let the person being asked to take over a school actually respond — accepting it, or turning it down.

## What's changed?

- `PUT /api/schools/:school_id/ownership_transfer/accept` — the nominee accepts. The transfer's status becomes "completed".
- `PUT /api/schools/:school_id/ownership_transfer/decline` — the nominee turns it down. The transfer's status becomes "rejected".

Both endpoints:
- Only work for the actual person who was nominated — not the owner who sent the request, and not anyone else at the school.
- Only work while the transfer is still waiting for a response. Trying to accept or decline something that's already been dealt with (or never existed) gets the same "not found" response either way, so no one can tell the difference between "there's nothing here" and "you're just not allowed to see it."
- Don't need anything in the request body — the school and the logged-in user are enough to know what's being responded to.

## Examples

Accepting or declining, as the nominee:

```
PUT /api/schools/9d9c1e3e-1e4b-4b0a-9a2e-6a2b7e6f1a10/ownership_transfer/accept
Authorization: Bearer <nominee's token>
```

```
200 OK
```

(empty body — the frontend already knows what happened because it made the request; it re-checks the transfer's status separately if it needs to)

If you're not the nominee, or there's nothing waiting to be responded to:

```
404 Not Found
```

Other responses: `401` (no token), `403` (you're a student — students were never allowed near this feature in the first place).


app/controllers/api/ownership_transfers_controller.rb
```diff
@@ -28,8 +28,28 @@ def create
...
+    end
+
     private
 
+    def resolve!(status)
+      transfer = pending_ownership_transfer
+
+      if transfer.blank? || cannot?(action_name.to_sym, transfer)
+        head :not_found
+      elsif transfer.update(status:)
```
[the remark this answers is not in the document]
DNR500: This will be covered on another issue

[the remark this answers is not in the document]
DNR500: Adressed

app/models/ability.rb
```diff
@@ -85,7 +88,7 @@ def define_school_owner_abilities(school:)
     can(%i[read create create_batch destroy], ClassStudent, school_class: { school: { id: school.id } })
     can(%i[read create destroy], :school_owner)
     can(%i[read create destroy], :school_teacher)
-    can(%i[read create], :ownership_transfer)
+    can(%i[read create accept decline], :ownership_transfer)
```
zetter-rpf: I'm surprised that owners can accept or decline a transfer, when the issues say that only teachers can accept. Do you know why is this here?

app/models/ability.rb
```diff
@@ -102,7 +105,7 @@ def define_school_teacher_abilities(user:, school:)
     can(%i[read create create_batch destroy], ClassStudent, school_class: { school: { id: school.id }, teachers: { teacher_id: user.id } })
     can(%i[read], :school_owner)
     can(%i[read], :school_teacher)
-    can(:read, :ownership_transfer)
+    can(%i[read accept decline], :ownership_transfer)
```
zetter-rpf: I'm not sure that this is doing what we expect. This is saying any teacher can accept an ownership transfer, no matter who the transfer is intended for.

This needs to verify that the current user is nominated user (and possibly that they are a member of the expected school).

I saw there is a check for nominated_user_id in the shared abilities but abilities in cancancan are additive so if that fails this will still pass.

app/models/ability.rb
```diff
@@ -42,6 +42,9 @@ def define_authenticated_abilities(user)
     can :read, OwnershipTransfer do |transfer|
       user.id == transfer.requested_by_user_id || user.id == transfer.nominated_user_id
     end
+    can %i[accept decline], OwnershipTransfer do |transfer|
+      user.id == transfer.nominated_user_id
+    end
```
zetter-rpf: Are we only allowing teachers to receive and accept ownership transfers? If so, why do both of these need to be in the shared ability file?

zetter-rpf (changes requested):
I think the permissions aren't setup right which makes this a risk to merge.

We should review the existing abilities around ownership transfer and introduce only what's needed for this. 

I would expect:

-  owners to have a read/create for a matching school
- teachers to have a read/accept/decline for a matching nominee (and possibly school)

app/controllers/api/ownership_transfers_controller.rb
```diff
@@ -28,8 +28,33 @@ def create
...
     private
 
+    # Wrapped in a transaction so the row lock below is held across the
+    # read-and-update, preventing a concurrent accept/decline on the same
+    # transfer from also finding it pending.
+    def resolve!(status)
+      OwnershipTransfer.transaction do
+        transfer = pending_ownership_transfer
+
+        if transfer.blank? || cannot?(action_name.to_sym, transfer)
```
zetter-rpf: is the cannot needed here? we normally authorize the resource when it's loaded which should automatically send the correct response.

DNR500: Yeah, I think it's needed to keep the API consistent - the normal auto-check would return a 403, which tells a non-nominee teacher a transfer exists. This gets us a 404 instead, same as `show`, so we don't leak that a pending transfer exists to someone who shouldn't know.

app/models/ability.rb
DNR500: Good catch - only teachers should accept/decline, so it shouldn't live in the shared file. Moved it into the teacher-only one: [fix: require the current user to be the nominee to accept or decline a transfer](https://github.com/RaspberryPiFoundation/editor-api/pull/1025/changes/2b8dc1ddca407cd062ee20490d729bb33d25a6cd). Read stays shared since both owner and nominee need to see it.

app/models/ability.rb
DNR500: Fixed in the same commit as the other ability.rb comment: [fix: require the current user to be the nominee to accept or decline a transfer](https://github.com/RaspberryPiFoundation/editor-api/pull/1025/changes/2b8dc1ddca407cd062ee20490d729bb33d25a6cd). It's not a blanket teacher permission anymore - accept/decline now only applies when the current user is the actual nominated user for that specific school.

app/models/ability.rb
DNR500: Yep I had missed this - owners were never meant to have that. Removed it here: [fix: school owners cannot accept or decline an ownership transfer](https://github.com/RaspberryPiFoundation/editor-api/pull/1025/changes/4a7b252b8c6532b164af2b2af950d8e15230e5bc). Owners are back to read/create only.

zetter-rpf (approved):
I talked to @DNR500 about this, good to be merged.

I have suggested that the accept/decline actions might be better in a new controller - they are dealing with the same model, but the resource is loaded differently and the authorisation is different too. Having it in different controllers can make this easier to follow and harder to accidentally use an unauthorised resource

app/controllers/api/ownership_transfer_resolutions_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    private
+
+    # Wrapped in a transaction so the row lock below is held across the
+    # read-and-update, preventing a concurrent accept/decline on the same
+    # transfer from also finding it pending. Authorizing against the loaded
+    # transfer (rather than checking cannot? by hand) means an unauthorized
+    # attempt raises and is rescued below into the same 404 a nonexistent
+    # transfer gets, instead of leaking that a pending transfer exists.
```
zetter-rpf, line 24: Not sure if this comment is giving much value.  It seems random to mention cannot given that there are many ways you could write this, this behaviour would be better handled by a test (which I think there already is one?)

The locking is harder to test, but also I think their intention is clear from looking at the code

zetter-rpf (approved):

